Issue #4, opened 2017-05-22
file watcher triggers too much


alexeagle:
If I carefully write to a file with a simple unix tool, like
`echo "" >> file` then I see exactly one rebuild when watching `file`.
But with any editor, after saving a file I see a bunch of rebuilds. On linux with VScode, I saw exactly two rebuilds, but on Mac it triggers endlessly and the `ibazel` command needs to be killed.

mprobst:
iblaze specifically debounces the writes for that reason, it triggers on the first fs event, but then collects fs events for another 100 ms (or so, not sure about the number). The reason is that most editors don't just write a file, they have some dance like writing a new copy, then move the copy over the old file, and sometimes even moving the old file to a backup before that, which of course causes multiple events.

achew22:
@alexeagle, is this happening on macOS? It could be additionally related to howeyc/fsnotify#62. The correct solution is to debounce this but while I look into how to do that it might be helpful to try their suggested fix of turning off spotlight indexing.

PR #7, opened 2017-05-24
Add a debounce of 100ms to ibazel actions.

achew22:
Also refactor so that there is no sleep at the end of iBazel. It will
now watch your files indefinitely.

Fixes: #4

### DIFF
--- a/ibazel/main.go
+++ b/ibazel/main.go
@@ -24,7 +24,20 @@ import (
 	"github.com/fsnotify/fsnotify"
 )
 
-var b *bazel.Bazel
+type State string
+
+const (
+	DEBOUNCE_QUERY State = "DEBOUNCE_QUERY"
+	QUERY          State = "QUERY"
+	WAIT           State = "WAIT"
+	DEBOUNCE_RUN   State = "DEBOUNCE_RUN"
+	RUN            State = "RUN"
+	QUIT           State = "QUIT"
+)
+
+const debounceDuration = 100 * time.Millisecond
+const sourceQuery = "kind('source file', deps(set(%s)))"
+const buildQuery = "buildfiles(deps(set(%s)))"
 
 func usage() {
 	fmt.Printf(`ibazel
@@ -53,32 +66,26 @@ func main() {
 		return
 	}
 
-	b = bazel.New()
-
 	command := os.Args[1]
 	target := os.Args[2]
 
-	query := fmt.Sprintf("kind('source file', deps('%s'))", target)
-
-	toWatch := queryForSourceFiles(query)
-
-	watcher, err := fsnotify.NewWatcher()
+	// Even though we are going to recreate this when the query happens, create
+	// the pointer we will use to refer to the watchers right now.
+	buildFileWatcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		fmt.Printf("Watcher error: %v", err)
 		return
 	}
-	defer watcher.Close()
+	defer buildFileWatcher.Close()
 
-	for _, line := range toWatch {
-		fmt.Printf("Line: %s\n", line)
-		err = watcher.Add(line)
-		if err != nil {
-			fmt.Printf("Error watching: %v", err)
-			return
-		}
+	sourceFileWatcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		fmt.Printf("Watcher error: %v", err)
+		return
 	}
+	defer sourceFileWatcher.Close()
 
-	var commandToRun func(*bazel.Bazel, string)
+	var commandToRun func(string)
 	switch command {
 	case "build":
 		fmt.Printf("Building %s\n", target)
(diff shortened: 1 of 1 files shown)


achew22:
@alexeagle, PTAL.

I don't have a mac, but I tested this manually by doing all sorts of nonsense with touching, moving and copying files. Does this fix your issue?

ibazel/main.go
```diff
@@ -94,27 +94,30 @@ func main() {
 		return
 	}
 
+	// Create a timer objet to use for debouncing. When the timer is created, do
+	// not start the timer but instead start the timer when an event that builds
+	// should be run.  This has the added benefit of kicking things off by
```
gconnell: not sure what you mean by "start the timer" here... NewTimer will start the timer upon creation.

ibazel/main.go
```diff
@@ -94,27 +94,30 @@ func main() {
...
-			select {
-			case _ = <-watcher.Events:
-				fmt.Printf("Files changed rebuilding...\n")
-				commandToRun(b, target)
-			case err := <-watcher.Errors:
-				fmt.Println("Error:", err)
-			}
+	for {
+		select {
+		case _ = <-watcher.Events:
```
gconnell: here and line 109, no need for the `_ =`.  Just say:  `case <-chan:`

ibazel/main.go
```diff
@@ -94,27 +94,30 @@ func main() {
...
+			fmt.Printf("Files changed...\n")
+			debouncer.Reset(debounceDuration)
```
gconnell: See discussion in https://golang.org/pkg/time/#Timer.Reset re: Stop and Reset.  I'm worried that you'll potentially have a full debouncer.C when calling Reset, in certain conditions where the watcher fires right when debouncer's time runs out, and you hit the first case instead of the second.

ibazel/main.go
```diff
@@ -94,27 +94,30 @@ func main() {
 		return
 	}
 
+	// Create a timer objet to use for debouncing. When the timer is created, do
+	// not start the timer but instead start the timer when an event that builds
+	// should be run.  This has the added benefit of kicking things off by
+	// sending an event to make the first run happen.
+	debouncer := time.NewTimer(debounceDuration)
```
gconnell: this debounce logic seems sound.  Here's another approach which might have slightly less complex code with slightly more variability of debounce time.  Your way works totally fine, though.

```go
tick := time.Tick(debounceTime / 2)
state := "RUN"
for {
  select {
    case <-event:
      state = "EVENT"
    case <-tick:
      switch state {
        case "EVENT":
          state = "RUN"
        case "RUN":
          runCommand()
          state = "DONE"
        case "DONE":
          // nothing
      }
  }
}

**Comment by alexeagle** [comment]
I tried re-building ibazel from this branch, now it doesn't find one of the inputs?

$ ibazel build src
Line: src/util.ts
Line: src/type-translator.ts
Line: src/tsickle_compiler_host.ts
Line: src/tsickle.ts
Line: src/source_map_utils.ts
Line: src/rewriter.ts
Line: src/modules_manifest.ts
Line: src/main.ts
Line: src/jsdoc.ts
Line: src/es5processor.ts
Line: src/decorators.ts
Line: src/decorator-annotator.ts
Line: src/cli_support.ts
Line: /tsconfig.json
Error watching: lstat /tsconfig.json: no such file or directory

Dunno why...

```
$ bazel query "kind('source file', deps('src'))" | tail
//src:source_map_utils.ts
//src:rewriter.ts
//src:modules_manifest.ts
//src:main.ts
//src:jsdoc.ts
//src:es5processor.ts
//src:decorators.ts
//src:decorator-annotator.ts
//src:cli_support.ts
//:tsconfig.json
```

**Comment by achew22** [comment]
@alexeagle can you try running a test to see if without this commit it still fails? I hypothesize that the way I'm going from query to filename is incorrect for files that exist in the root of the workspace. Specifically the problem with tsconfig.json is that the target is "//:tsconfig.json". When it replaces the ":" with a "/" the filename is describing something at the root of the filesystem. As a simple fix for this I wonder if I can just prepend "./" to the path when I watch it.

Does this pass the sniff test? Could you try making line 132 of ibazel/main.go be 

```
toWatch = append(toWatch, "./" + strings.Replace(strings.TrimPrefix(line, "//"), ":", "/", 1))
```

and report back if that fixes it?

File: ibazel/main.go [hunk 5]
```diff
@@ -53,30 +66,16 @@ func main() {
...
 	}
 
-	b = bazel.New()
+	b := bazel.New()
 
 	command := os.Args[1]
 	target := os.Args[2]
 
-	query := fmt.Sprintf("kind('source file', deps('%s'))", target)
+	var buildFileWatcher *fsnotify.Watcher
```
mprobst: One problem in my experience is that `*fsnotify.Watcher.Events` is not buffered. That means it'll block once some event happens, and the interleaving of you reading from the channel and events arriving is complex and can lead to bouncy behaviour (if not races).

What you can is create your own channel with appropriate buffering (10 items or so) and a goroutine that continuously reads fsnotify's channel and writes into the buffer.

Then to debounce, the simplest approach I can think of is to just `time.Sleep(100 * time.Millisecond)` (or whatever number, 20 ms might be enough, too), then synchronously read all events of the buffered channel, i.e. deplete it. That gives a relatively straight forward algorithm that should cover most eventualities (unless you e.g. have a ton of events arriving).

achew22:
@gconnell, could you take another look at how this is structured? I would appreciate it.

ibazel/main.go
```diff
@@ -24,7 +24,20 @@ import (
...
+	QUERY          State = "QUERY"
+	WAIT           State = "WAIT"
+	DEBOUNCE_RUN   State = "DEBOUNCE_RUN"
+	RUN            State = "RUN"
+	QUIT           State = "QUIT"
+)
+
+const debounceDuration = 100 * time.Millisecond
+const sourceQuery = "kind('source file', deps('%s'))"
+const buildQuery = `let d = deps(set(%s)) in buildfiles($d)`
```
mprobst: I think you can just write `buildfiles(deps(set(%s)))`, no need for the variable.

ibazel/main.go
```diff
@@ -24,7 +24,20 @@ import (
...
+	DEBOUNCE_QUERY State = "DEBOUNCE_QUERY"
+	QUERY          State = "QUERY"
+	WAIT           State = "WAIT"
+	DEBOUNCE_RUN   State = "DEBOUNCE_RUN"
+	RUN            State = "RUN"
+	QUIT           State = "QUIT"
+)
+
+const debounceDuration = 100 * time.Millisecond
+const sourceQuery = "kind('source file', deps('%s'))"
```
mprobst: should this also use `set`, in case the user queries multiple targets?

ibazel/main.go
```diff
@@ -94,27 +101,54 @@ func main() {
...
+			case <-sourceFileWatcher.Events:
+				fmt.Printf("Detected source change. Rebuilding...\n")
+				state = DEBOUNCE_RUN
+			case <-buildFileWatcher.Events:
+				fmt.Printf("Detected build graph change. Requerying...\n")
+				state = DEBOUNCE_QUERY
+			}
+		case DEBOUNCE_QUERY:
+			state = QUERY
+			select {
```
mprobst: would this be more readable as a:

```
for {
  select {
  case <-time.After(debounceDuration):
    break
  case <-buildFileWatcher.Events:
    continue
  }
}
```

I find it a bit confusing that it switches back and forth between query and debounce here, and continues the outer loop.

ibazel/main.go
achew22: Done

ibazel/main.go
achew22: Done

ibazel/main.go
achew22: I am going to make a wrapper around the watchers that denounces for me in the next CL and remove the debounce states.

achew22:
I don't think anything in here is too offensive now. I'm going to merge this so I can start working on refactoring this some.

Issue #9, opened 2017-05-31
Track files who are replaced (by mv operations) when being saved


achew22:
Some editors writes files by writing them to a file then moving them to replace the file they are writing. I presume this is for atomicity reasons on some file systems. Since the inode changes files are occasionally lost by ibazel which prevents it from working when you write the file the second time.

To fix this we can readd a watch whenever a file is moved, replacing the original file.

PR #10, opened 2017-05-31
Track files replaced by move

dougkoch:
Adds a watch back to any file that has been replaced by a move. We look for a fsnotify.Remove event and retry watcher.Add() until it returns an error code of 0. Once a max number of retries has been exceeded, it is assumed that the file was removed.

Fixes: #9

### DIFF
--- a/ibazel/main.go
+++ b/ibazel/main.go
@@ -59,25 +59,6 @@ ibazel build //path/to/my/buildable:target
 `)
 }
 
-type SourceEventProcessor struct {
-	SourceFileEvents chan fsnotify.Event
-	SourceFileWatcher *fsnotify.Watcher
-}
-
-func (s SourceEventProcessor) Listen() {
-	for {
-		select {
-		case event := <-s.SourceFileWatcher.Events:
-			s.SourceFileEvents <- event
-
-			switch event.Op {
-			case fsnotify.Remove:
-				s.SourceFileWatcher.Add(event.Name)
-			}
-		}
-	}
-}
-
 func main() {
 
 	if len(os.Args) < 3 {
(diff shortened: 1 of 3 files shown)


bazel-io:

mprobst:
Alternative approach: run the build, then drop and re-add all watches. This usually gives enough time for the file to re-appear.

dougkoch:
@mprobst Turns out that a file overwritten by a mv re-appears quickly enough that a watch can be added as soon as the remove event is received. My retry loop (removed in my latest commit) was unnecessary.
@achew22 Ready for review

ibazel/main.go
```diff
@@ -59,6 +59,25 @@ ibazel build //path/to/my/buildable:target
 `)
 }
 
+type SourceEventProcessor struct {
```
achew22: Can you move this into its own file?

ibazel/main.go
```diff
@@ -59,6 +59,25 @@ ibazel build //path/to/my/buildable:target
...
+	SourceFileEvents chan fsnotify.Event
+	SourceFileWatcher *fsnotify.Watcher
+}
+
+func (s SourceEventProcessor) Listen() {
```
achew22: Can you add a method that creates a SourceEventProcessor and creates the channel then calls

go s.Listen()

ibazel/main.go
```diff
@@ -107,7 +132,7 @@ func main() {
 		switch state {
 		case WAIT:
 			select {
-			case <-sourceFileWatcher.Events:
+			case <-sourceFileEvents:
```
achew22: Can you refer to the channel as being attached to the object instead of in the function scope?

ibazel/main.go
dougkoch: Done

ibazel/main.go
dougkoch: Done

ibazel/main.go
dougkoch: Done

Issue #12, opened 2017-06-01
ibazel doesn't support multiple target patterns even though it is documented to do so


endobson:
```
Usage:

ibazel build|test|run targets...
```

yet from `main.go`
```
	target := os.Args[2]
```

PR #14, opened 2017-06-07
Support multiple target patterns

dougkoch:
As @endobson pointed out, the CLI usage message implies that multiple target patterns are supported, even though they're not.

Fixes: #12


bazel-io:

dougkoch:
@achew22 I updated my PR to do multiple targets in one command; the way to make it work is to pass them as varargs, e.g. `b.Build(targets...)`

PR #16, opened 2017-06-28
Refactor main.go into testable chunks

achew22:
The iBazel object performs the same as the main function did before but
is broken into smaller, tested parts.


achew22:
Retest this please

achew22:
Looks like head is broken but it works on 0.5.2 so I am going to commit

Issue #15, opened 2017-06-13
publish gh release binaries


softprops:
something like https://github.com/bazelbuild/buildtools/releases would be ideal. using a naming convention what would allow for conditional download based on platform using standard `uname` conventions would be even better

achew22:
Do you have any idea how those are generated? I wonder if I can piggyback on their packaging strategy

ittaiz:
Not sure it's better. I don't object to this but the release cycle of Bazel
can take time which you often don't want to wait for ibazel

alexeagle:
Fair enough, we should keep an independent release process until ibazel is
stable.

alexeagle:
@achew22 how about if we also publish ibazel on NPM? That's the most convenient way to distribute it to Angular users.
/cc @vsavkin

achew22:
100% happy to do that. Do you have a link to docs on how to do that?

alexeagle:
I can set it up and document. Big part is building for each arch on each
release. @mprobst does that for clang-format

achew22:
I sent an email to the Bazel release manager internally so I might try to
piggy back off of their work.

On the plus side, rules_go just added CROSSTOOL support so it shouldn't be
that painful.

ittaiz:
Bazel uses brew and apt-get right?
That is really comfortable (though I imagine more work for the maintainer
than npm)

softprops:
👋 hi, opener of the original issue here. I think the idea of having support for change triggered actions built into bazel would be ideal but for my company that issue becomes a liability in turn around time. We like fast turn arounds when blocked. Thats what ended up facilitating our work on a separate tool https://github.com/meetup/rebazel. I'd worry that having this built into bazels release cycles would not make the turn around process any faster and I'm fact very much slower

ittaiz:
Andrew,
Do you mean to piggy back off their release cycle or their release infra to
have your own releases?

achew22:
@softprops && @ittaiz,

The objective would be to use the CI servers for the heavy lifting of secret storage, scheduling, and uploading. That will make it trivial to actually do releases which I've found makes releases happen more often.

I've put together my requirements/wish list on bazelbuild/continuous-integration#100. Feel free to follow along there. I will keep the conversation as public as I can.

Globegitter:
[...]

PR #24, opened 2017-08-27
Prepare to publish to NPM:

alexeagle:
Include a binary distribution in the package (only Mac for now)
A wrapper script to fork the native process

Note: the version field on NPM doesn't correspond to anything right now,
I think we should at least tag this repo with a matching version for sanity
We also need some README about performing releases, since no one will remember
when to re-build the binary.

See #15


achew22:
Alex, just got back from my vacation. This totally looks like the way to go for publishing to NPM. My only hesitation is toward checking in the binaries. Would it bother you if I refer to the `ibazel` binary in the bazel-bin directory and wrapped the uploading in a small bash script?

Do you have any suggestions on owning/managing NPM organizations? Is that the way to go or should I just set it up under a personal account for now?

alexeagle:
"wrapped the uploading" - npm packages are published with the `npm publish` command which expects a directory that contains the artifact tree. We could have a build process that creates such a directory, which would need to be populated with binaries for more platforms over time.

I already have the keys to http://npmjs.com/org/bazel for when we're ready to publish it.

achew22:
Excellent! I sat down and wrote a wrapper for this tonight based on your PR so that we can start publishing to NPM. Do you know what I need to do to make a bazel/ibazel npm package that can be uploaded to?

alexeagle:
`npm login`
tell me your username so I can grant permissions
`npm publish <folder containing package.json>`

achew22:
I think this is probably obviated by #44

achew22:
Closing since #44 landed. Thank you for doing all the hard work on this one!

PR #27, opened 2017-09-06
Kill preexisting when ibazel run detects changes

achew22:
It feels silly that I didn't catch this sooner cause it is obvious in
hindsight. When you run `bazel run` on a task it blocks usage of Bazel
until the task finishes. Normally this isn't a problem, but the most
useful use of `bazel run` is when working with servers that are long
lived and need to be restarted. Since Bazel was locked by the naive
`bazel run` call, ibazel was never able to get the lock and actually run
the new code thus completely defeting the purpose of ibazel.

Now, we run `bazel build` and then start the subprocess, tracking the
PGID, and send a kill to the PGID when a change has been detected and a
rebuild is going to occur. This should make `ibazel run` useful in the
run mode.

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -1,8 +1,26 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


mprobst:
You can actually run the binary from bazel-bin without blocking the bazel
server. I found that to be more useful than respawning the binary. In
particular for servers that often take a long time to start, it's useful to
rebuild their assets (think: JS, TypeScript, etc) without restarting the
server.

achew22:
@mprobst, I believe we are talking about the same thing. in the PR, I `bazel build` the target specified by the user and then I exec the binary in the bazel-bin dir.

It's an interesting idea for when you're doing frontend work to only rebuild the data dependencies for the server. Can we continue this conversation in #29?

WRT restarting servers that take a long time to start (I assume this is mostly Java) I have filed #28 to track. I'm going to also merge this because I have another PR that I would like to finish up that does argument parsing which is a long standing feature request but it depends on this code being merged.

mprobst:
@achew22 I see, I replied just based on your PR description without looking at the code.

PR #30, opened 2017-09-07
Add a basic argument processor.

### LINKED ISSUE #13: ibazel does not pass arguments to Bazel
I want to run my tests in parallel and ibazel adds `--test_output=streamed` to the args when running `bazel test`, which is unexpected and undocumented. I would expect it to just run exactly the same had the command line been `bazel test`.

achew22:
Now when you pass in arguments to ibazel it should pass them through to
bazel or to your running job if you pass it after a `--`.  Fixes #13.

Also this no longer sets --test_output=streamed by default.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -23,6 +23,7 @@ import (
 )
 
 type Bazel interface {
+	SetArguments([]string)
 	WriteToStderr(v bool)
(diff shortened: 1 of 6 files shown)


ittaiz:
Exciting, thanks!

PR #33, opened 2017-09-08
wait for process to end when killing it

jmhodges:


bazel-io:

achew22:
I'm not sure how to do this so feel free to push back and say "that's hard and I don't know how to do it either", but is it possible to add a test for this? Are there any standard linux commands (i.e. shipped in busybox) that would exhibit this? I would guess not but I figured I'd ask.

Also -- Jenkins, test this please.

jmhodges:
They'd all be really racy tests. I don't know of any utilities that would reliably delay across OSes. I could do it if we sent SIGTERM instead of SIGKILL

achew22:
I figured. Thanks for the patch!

achew22:
Looks like Travis is having issues with their Mac builds. I'm just going to merge since I can't get the CI to even run for it on Mac.

https://www.traviscistatus.com/incidents/4qylrqvy50gy

jmhodges:
Thank you!

PR #34, opened 2017-09-09
Use bazel run --script_path

achew22:
Bazel run provides a nice helper --script_path that correctly sets up
the working dir and related then launches the program in question. This
should prevent the error where you're in the wrong directory or missing
an environment variable when running ibazel run but not bazel run since
we now invoke in the same way.


achew22:
Travis appears to still be broken but jenkins approves so LGTM

PR #35, opened 2017-09-10
Listen for SIGINT/KILL and quit.

achew22:
Listen for SIGINT and kill the subprocess. If it doesn't respond in 3
whacks, exit ibazel.

If it gets a SIGKILL, attempt to kill the subprocess and then exit.

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


achew22:
Travis appears to still be backlogged so I'm going to submit now that Jenkins approves.

jmhodges:
This is racy, no? `i.cmd` is accessed in both goroutines?

jmhodges:
I think there's a also a non-memory race but still a race even after adding some locks where the check in `subprocessRunning` can return false and then before the exit is called, a new process can be started up. That's one reason why justrun had to pick up the cmdReloader https://github.com/jmhodges/justrun/commit/97d05ac9e7cd40fedf1f2beaea3066af98ae1436

PR #39, opened 2017-09-28
Create NotifyCommand mode

achew22:
In this mode the command is notified that the build has completed over
stdin. The program can then intelligently react and decide to reload
dependencies without having to restart.

### DIFF
--- a/ibazel/command/notify_command_test.go
+++ b/ibazel/command/notify_command_test.go
@@ -1,0 +1,87 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"errors"
+	"os/exec"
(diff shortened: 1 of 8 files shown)


ibazel/command/notify_command_test.go
```diff
@@ -0,0 +1,87 @@
...
+	c.stdin, err = cmd.StdinPipe()
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Mock out bazel to return non-error on test
+	b := &mock_bazel.MockBazel{}
+	b.WaitError(nil)
+	bazelNew = func() bazel.Bazel { return b }
+	defer func() { bazelNew = oldBazelNew }()
```
mrmeku: This might be might lack of go skill talking but it seems like these two lines are the equivelent of writing bazelNew = oldBazelNew. Why defer this function?

ibazel/command/notify_command_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	c.NotifyOfChanges()
+	b.WaitError(errors.New("Demo error"))
+	c.NotifyOfChanges()
+	b.WaitError(nil)
+	c.NotifyOfChanges()
+
```
mrmeku: Might be good to expect that isSubprocessRunning to be true at this point

ibazel/command/notify_command_test.go
achew22: Defer executes a function at the end of the current function. This makes the bazelNew mock get reset after everything is done.

ibazel/command/notify_command_test.go
achew22: Done

achew22: Huh, turns out that isn't possible to do because I'm using `cmd.CombinedOutput` which starts the job. I'm going to return it to the old way and work instead on an end to end test which is a better test anyway.

PR #41, opened 2017-09-29
Detect the tag iblaze_notify_changes and use NotifyCommand instead

achew22:


ibazel/BUILD
```diff
@@ -18,6 +18,9 @@ go_binary(
     name = "ibazel",
     importpath = "github.com/bazelbuild/bazel-watcher/ibazel",
     library = ":go_default_library",
+    tags = [
+        "IBAZEL_MAGIC_TAG",
```
mrmeku: Why is this here? Can you leave a comment saying why this is necessary?

third_party/bazel/master/src/main/protobuf/build.proto
```diff
@@ -0,0 +1,569 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// This file contains the protocol buffer representation of a build
+// file or 'blaze query --output=proto' call.
+
```
mrmeku: Do you want to say where this proto was sourced from?

achew22: Done and added some more tests

PR #44, opened 2017-10-03
Create script for generating npm releases

achew22:
Tested by publishing in a temp namespace https://www.npmjs.com/package/@achew22/ibazel

@alexeagle, could you try this out?

### DIFF
--- a/npm/index.js
+++ b/npm/index.js
@@ -1,0 +1,48 @@
+#!/usr/bin/env node
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict';
+
+// This package inspired by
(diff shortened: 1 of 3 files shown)


npm/package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "@bazel/ibazel",
+    "version": "0.0.1",
```
alexeagle: how will this number be updated as we tag versions in git?

npm/index.js
```diff
@@ -0,0 +1,47 @@
...
+// https://github.com/angular/clang-format/blob/master/index.js
+const os = require('os');
+const path = require('path');
+const spawn = require('child_process').spawn;
+
+function main(args) {
+  const arch = {
+    'x64' : 'amd64',
+  }[os.arch()];
+  const platform = {
```
alexeagle: not just const platform = os.platform() ?

npm/package.json
achew22: Right now I don't have any git tags but it will be easy to sed it to the git tag when we do that. NPM will prevent publishing multiple versions with the same number so I believe this hardcoded is safe for the moment

npm/README.md
```diff
@@ -0,0 +1,38 @@
+# Bazel watcher
```
alexeagle: can you avoid checking in a copy of the file? just change publish.sh to get the README from / instead?

achew22: Rm'd

npm/package.json
achew22: I changed it to v0.0.1 so that it will match what I intend to tag in git with

npm/index.js
achew22: If I use os.Platform directly and someone is on solaris (who knows why) then the file won't exist and they won't error out.

PR #46, opened 2017-10-04
Clean up the logging for successful watched files

### LINKED ISSUE #40: Don't write all the watched file locations
In my build graph, there are hundreds of source files transitively depended on by the target I `iblaze`

Printing the list of files is spamming my terminal. Also, I'm not sure how this is useful to an end user, if they need to debug what is being watched you could just print an `ibazel query ...` command that would show them.

Perhaps if you do think it's useful to print the list, it should be a CLI option? (but I think increasing the CLI surface should be done sparingly)

mrmeku:
Before this change a line was printed for every single file that was successfully watched.
After this change one line is printed stating the total count of watched files.

Fixes #40


bazel-io:

achew22:
Test this please

PR #48, opened 2017-10-07
Add an end to end test

achew22:


ittaiz:
Hi,
I hope I'm not barging in but have you considered using the new published Bazel e2e support? It's a new repo under Bazelbuild which allows to depend and run Bazel itself in e2e. It currently has python and java support but I'm sure they'd love go support.

achew22:
Yes, I did consider that, but I will probably not be using the `blaze` CLI as the primary interface for reading `BUILD` files for much longer so I put together something quick that did what I needed and nothing more. Maybe way down the road it will be worth testing but ATM it seems like a lot more work with little benefit. My goal is to ultimately not need to invoke the Bazel CLI from iBazel.

ittaiz:
Interesting and a bit scary altogether :)
I think part of the (mistaken?) appeal of ibazel to me is that it's a thin
wrapper on top of calling Bazel directly.
If it separates from it then won't this introduce all kinds of situations
like "I wonder why my CL broke, it passed using ibazel locally"?
Also,
Is there a different issue you'd rather we use for this discussion?
It sounds important

ibazel/main.go
```diff
@@ -81,13 +84,17 @@ func parseArgs(in []string) (targets, bazelArgs, args []string) {
...
+	if len(flag.Args()) < 2 {
 		usage()
 		return
 	}
 
-	command := strings.ToLower(os.Args[1])
-	args := os.Args[2:]
+	command := strings.ToLower(flag.Args()[0])
+	args := flag.Args()[1:]
+	fmt.Printf("Command: %s Args: %v", command, args)
```
mrmeku: Is this print statement worth keeping around? In most use cases this information is present from what the user just typed. I think it'll look something like

ibazel run //path/to:target
Command: run Args: //path/to:target

ibazel run //path/to:target --bazelPath=/path/to/bazel
Command: run Args: //path/to:target --bazelPath=/path/to/bazel

.travis.yml
```diff
@@ -11,7 +11,7 @@ os:
   - osx
 
 env:
-  - V=0.5.4
+  - V=0.6.1
```
mrmeku: 0.6.0?

e2e/bazel/bazel.go
```diff
@@ -0,0 +1,180 @@
...
+			fmt.Errorf("I have no idea what a %T is... please fix.", v)
+			panic(v)
+		}
+	}
+
+	return true
+}
+
+func runWithScript(args []string) {
+	// Since all ru ncommands start with "run" as arg0 and it breaks flag
```
mrmeku: // All run commands start with "run" as the first argument while all other arguments are flags.

e2e/bazel/bazel.go
```diff
@@ -0,0 +1,180 @@
...
+	// parsing, strip args[1]. If you need it, imagine there is a "run" at the
+	// beginning.
+	args = args[1:]
```
mrmeku: flags := args[1:]

e2e/bazel/bazel.go
```diff
@@ -0,0 +1,180 @@
...
+
+	fs := flag.NewFlagSet("bazel_run_flags", flag.ExitOnError)
+	scriptPath := fs.String("script_path", "default", "")
+	err := fs.Parse(args)
```
mrmeku: err := fs.Parse(flags)

ibazel/ibazel.go
```diff
@@ -214,7 +214,7 @@ func (i *IBazel) iteration(command string, commandToRun func(...string), targets
 		}
 	case RUN:
 		i.state = WAIT
-		fmt.Printf("%sing %s\n", strings.Title(command), joinedTargets)
+		fmt.Fprintf(os.Stderr, "%sing %s\n", strings.Title(command), joinedTargets)
```
mrmeku: Should this and other print statements in the QUERY stage be logged to Stdout?

ibazel/command/default_command.go
```diff
@@ -64,9 +65,9 @@ func (c *defaultCommand) Start() {
 
 	var err error
 	if err = c.cmd.Start(); err != nil {
-		fmt.Printf("Error starting process: %v\n", err)
+		fmt.Fprintf(os.Stderr, "Error starting process: %v\n", err)
 	}
-	fmt.Printf("Starting...")
+	fmt.Fprintf(os.Stderr, "Starting...")
```
mrmeku: Stdout?

e2e/simple_test.go
```diff
@@ -0,0 +1,62 @@
...
+		pids = append(pids, pid)
+
+		expectedOut := ""
+		for i := 0; i < count; i++ {
+			expectedOut += "Started!"
+		}
+		assertEqual(t, expectedOut, ibazel.GetOutput(), "Ouput was inequal")
+
+		expectedErr := ""
+		for i := 0; i < count; i++ {
```
mrmeku: This for loop is really cool and all, but it doesn't do anything.

e2e/common.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var bazelPath string
+var ibazelPath string
+
+const mainGoPath = "e2e/simple/main.go"
+const BUILDPath = "e2e/simple/BUILD.bazel"
+
+func init() {
+	bazelPath = getPath("e2e/bazel/bazel")
```
mrmeku: Is there a reason you don't initialize these variables before this function is run? If so leave a comment

e2e/bazel/bazel.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+var sourceFiles = &blaze_query.QueryResult{
+	Target: []*blaze_query.Target{
+		&blaze_query.Target{
+			Type: blaze_query.Target_SOURCE_FILE.Enum(),
+			SourceFile: &blaze_query.SourceFile{
+				Name:            proto.String("//e2e/simple:main.go"),
+				Location:        proto.String("/home/user/go/src/github.com/bazelbuild/bazel-watcher/e2e/simple/BUILD.bazel:3:1"),
```
mrmeku: Is the prefix /home/user/go/src/ part of your own system set up? Is it arbitrary? It seemed out of place so I just wanted to check with you about it

e2e/bazel/bazel.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	err = sp.Chmod(0755)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Fprintf(sp, `#! /usr/bin/env bash
```
mrmeku: Move this into a helper function with a descriptive name. Its a bit hard to transition from reading beautiful go code to a *insert your own adjective to describe bash syntax* bash script

ibazel/command/default_command.go
achew22: since stdout/err are combined when displayed on screen, my plan is to reserve stderr for the communication to the user from `ibazel` leaving the stdout pipe free for your program to output stuff. This is in the same tradition as `bazel` proper which has this output (when you pipe stdout to /dev/null and totally ignore it)

```
achew22@lappy:~/.../bazel-watcher (e2e)$ bazel build //... > /dev/null
.........
Loading: 
Loading: 0 packages loaded
Analyzing: 19 targets (9 packages loaded)
Analyzing: 19 targets (16 packages loaded)
WARNING: <snip>/WORKSPACE:1: Workspace name in <snip>/WORKSPACE (@io_bazel_rules_go) does not match the name given in the repository's definition (@com_github_bazelbuild_rules_go); this will cause a build error in future versions
Analyzing: 19 targets (32 packages loaded)
INFO: Analysed 19 targets (37 packages loaded).
INFO: Found 19 targets...
[1 / 6] BazelWorkspaceStatusAction stable-status.txt
[170 / 180] no action running
INFO: Elapsed time: 6.782s, Critical Path: 0.24s
INFO: Build completed successfully, 1 total action
```
So all the things that aren't explicitly important to your build go in the stderr pipe.

It turns out this functionality is really useful because it lets you pipe the output of your program to a file and still see build progress and stuff.

ibazel/ibazel.go
achew22: See my above comment.

achew22:
Bazel is doing the thing where it passes all the tests but `java.lang.IllegalStateException`s the results. I'm going to submit. Thanks for the review, @mrmeku.

Issue #57, opened 2017-11-01
Communicate to a long-running process that it will get stdin notifications


alexeagle:
There should also be an environment variable so the long-running process knows that it will get input on stdin.
Otherwise the program cannot know if it was launched under `bazel` or `ibazel`, if it expects the latter, then under bazel it will hang waiting for stdin that never arrives

```
if (process.env.IBAZEL_NOTIFY_CHANGES == 'y') {
  // rely on stdin to find out about changes
} else {
  // do our own file watching, perhaps
}
```

Issue #56, opened 2017-11-01
ibazel writes IBAZEL_BUILD_COMPLETED FAILURE to stdin even for successful builds


alexeagle:
In my prototype devserver that uses ibazel from HEAD, I had to listen for FAILURE
https://github.com/alexeagle/angular-bazel-example/blob/98af46d27bab378e4ed5f14799c80578f0287d55/airplane/devserver.js#L9
because ibazel never reports SUCCESS

mrmeku:
I'm going to document a bit of research I've done on this bug. It appears that the "error" that ibazel identifies is one is causes it self. To see if an error is produced ibazel calls the .Wait method on the Cmd object it uses to represent an invocation of bazel. However, when it calls Wait to see if there was an error the error produced reads "Wait was already called". So I think the key to fixing this bug is to ensure that wait is only called once. If we do that then bazel should go down the code path which logs IBAZEL_BUILD_COMPLETED SUCCESS

Issue #55, opened 2017-11-01
Replace the "MAGIC" tag with something permanent; document it


alexeagle:
https://github.com/bazelbuild/bazel-watcher/blob/35e574c96178c8bc446c5052fde477f0a3866425/ibazel/ibazel.go#L265

@achew22 says the term `IBAZEL_MAGIC_TAG` was meant as a placeholder. I want to start depending on this feature so we should mature it to something we can live with, and document how it's used.

achew22:
Not closing since it isn't documented.

PR #60, opened 2017-11-11
Notify command now watches on the correct tag.

achew22:
Additoinally it emits the correct signals, stdin (from the perspective
of the code living in the repo being built), and environment variables.

Fixes: #57
Fixes: #56
Issue: #55

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -194,7 +194,13 @@ func (b *bazel) Run(args ...string) (*exec.Cmd, error) {
 }
 
 func (b *bazel) Wait() error {
-	return b.cmd.Wait()
+	res := b.cmd.Wait()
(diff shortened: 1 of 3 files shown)


ibazel/command/notify_command_test.go
```diff
@@ -69,11 +69,6 @@ func TestNotifyCommand(t *testing.T) {
 		[]string{"Build", "//path/to:target"},
 	})
 
-	err = c.stdin.Close()
```
dougkoch: Why are you no longer closing the stdin pipe after writing to it? This diff is what's causing the test to time out and fail; the notify command is blocking on the read.

achew22: Fixed

dougkoch:
LGTM

Issue #54, opened 2017-10-26
query fails quietly, easy to overlook and hard to debug


rkirov:
When running ibazel on the tsickle repro, I get:

```
State: QUERY
Querying for BUILD files...
Error running Bazel exit status 2
```

And goes on to do a single run, making it easy to overlook.

I can give you a better repro, or maybe error reporting could be improved between ibazel and bazel to report the query error.

Issue #2, opened 2017-05-19
When initial build fails, ibazel should exit


alexeagle:
$ ibazel build //example:all
Error running Bazel exit status 7
Building //example:all
Files changed rebuilding...
ERROR: while parsing '//example:all': no such package 'example': BUILD file not found on package path.
INFO: Elapsed time: 0.182s
Build error: exit status 1     
[ tool is now apparently hung ]

I think it's better to exit in this case (like iblaze does)

achew22:
I've several times found myself writing `while true; do iblaze build $target; done` in order to try and fix up some vexing syntax errors (usually error prone yelling at me :wink:). What would you think about bailing if the query fails but not if the build/test/run fails?

alexeagle:
True, I've also found the asymmetry of bailing out after a failed build annoying. @mprobst do you have a strong case for that behavior?

mprobst:
I think this should only fail if the query fails (i.e. if bazel-watcher doesn't know what to watch).

alexeagle:
Here's a confusing case: I do
`ibazel run :devserver`
and there is a compile error. My devserver is not started. But, ibazel is still running.
Now I fix the compile error, ibazel reports success. But the devserver is still not running.

It feels like either ibazel should exit if first build is not successful, or else `ibazel run` should have a special case that it starts the binary on the first successful build.

PR #63, opened 2017-11-13
Exit when query fails and pass through its stderr

dougkoch:
From the discussion on #2, it seems like the consensus was that ibazel should exit if the query fails (but not if the build fails).

Fixes: #2
Fixes: #54

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -284,8 +284,6 @@ func (i *IBazel) run(targets ...string) {
 
 func (i *IBazel) queryRule(rule string) (*blaze_query.Rule, error) {
 	b := i.newBazel()
-	b.WriteToStderr(false)
-	b.WriteToStdout(false)
 
 	res, err := b.Query(rule)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


bazel-io:

achew22:
Jenkins, please test this

achew22:
Jenkins, test this please

ibazel/ibazel.go
```diff
@@ -310,7 +310,7 @@ func (i *IBazel) queryForSourceFiles(query string) []string {
 	res, err := b.Query(query)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error running Bazel %s\n", err)
-		return []string{}
+		osExit(1)
```
achew22: WDYT about doing

```
	b := i.newBazel()
	b.WriteToStderr(true)
	b.WriteToStdout(true)

	res, err := b.Query(query)
```

here? It will return almost instantaneously and print the error to the UI which would be less terrible than parsing it out of the binarypb.

dougkoch: I'm not parsing the error out of the pb. Only the query's stdout is a pb message (containing the QueryResult), whereas its stderr is just a string. I'm just collecting that string and passing it through to ibazel's stderr. In light of that, doesn't my approach make more sense than running the same query a 2nd time?

PR #69, opened 2017-11-15
Live reload

gregmagolan:
- Adds off-the-shelf live reload server (github.com/jaschaephraim/lrserver) that is turned on via `ibazel_live_reload` tag in run target.
- Live reload server defaults to port 35729 but ports are scanned and first available port between 35729 and 35828 (range of 100) is chosen
- Adds `-nolive_reload` flag which forces live reload server off even if `ibazel_live_reload` is set
- Adds `-profile_dev` flag which turns on profile (just console logs at the moment)
- `ibazel_live_reload` flag usage requires fix to rules_typescript on my fork: https://github.com/gregmagolan/rules_typescript/tree/devserver

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -28,7 +28,7 @@ import (
 	"github.com/bazelbuild/bazel-watcher/bazel"
 	"github.com/bazelbuild/bazel-watcher/ibazel/command"
 	"github.com/fsnotify/fsnotify"
-	"github.com/jaschaephraim/lrserver"
+	"github.com/gregmagolan/lrserver"
 
 	blaze_query "github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf"
 )
@@ -153,6 +153,9 @@ func (i *IBazel) SetDebounceDuration(debounceDuration time.Duration) {
 func (i *IBazel) Cleanup() {
 	i.buildFileWatcher.Close()
 	i.sourceFileWatcher.Close()
+	if i.lrserver != nil {
+		i.lrserver.Close()
+	}
 }
 
 func (i *IBazel) setup() error {
@@ -274,7 +277,7 @@ func (i *IBazel) build(targets ...string) {
 	b.WriteToStdout(true)
 	err := b.Build(targets...)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Build error: %v", err)
+		fmt.Fprintf(os.Stderr, "Build error: %v\n", err)
 		return
 	}
 }
@@ -287,7 +290,7 @@ func (i *IBazel) test(targets ...string) {
 	b.WriteToStdout(true)
 	err := b.Test(targets...)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Build error: %v", err)
+		fmt.Fprintf(os.Stderr, "Build error: %v\n", err)
 		return
 	}
 }
@@ -306,7 +309,12 @@ func (i *IBazel) startLiveReloadServer() {
 	for ; port < lrserver.DefaultPort + 100; port++ {
 		if (testPort(port)) {
 			i.lrserver = lrserver.New("live reload", port)
-			go i.lrserver.ListenAndServe()
+			go func() {
+					err := i.lrserver.ListenAndServe()
+					if err != nil {
+							fmt.Fprintf(os.Stderr, "Live reload server failed to start: %v\n", err)
+					}
+			}()
 			url := fmt.Sprintf("http://localhost:%d/livereload.js?snipver=1", port)
 			os.Setenv("IBAZEL_LIVERELOAD_URL", url)
 			return
@@ -318,7 +326,7 @@ func (i *IBazel) startLiveReloadServer() {
 func (i *IBazel) setupRun(target string) {
 	rule, err := i.queryRule(target)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %s", err)
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
 	}
 
 	commandNotify := false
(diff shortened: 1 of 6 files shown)


bazel-io:

gregmagolan:
@alexeagle live reload PR

achew22:
Test this please.

@gregmagolan Thanks for you contribution. Lemme look at this really quick and send some feedback.

ibazel/ibazel.go
```diff
@@ -128,13 +136,22 @@ func (i *IBazel) SetBazelArgs(args []string) {
...
+func (i *IBazel) SetNoLiveReload(noLiveReload bool) {
+	i.noLiveReload = noLiveReload
+}
+
+func (i *IBazel) SetProfileDev(profileDev bool) {
+	i.profileDev = profileDev
+}
+
 func (i *IBazel) Cleanup() {
 	i.buildFileWatcher.Close()
```
achew22: It seems like there should be a step where the server gets shut down in here. If not, can you help me understand why?

ibazel/ibazel.go
```diff
@@ -57,16 +60,20 @@ type IBazel struct {
 	cmd       command.Command
 	args      []string
 	bazelArgs []string
+	noLiveReload bool
```
achew22: Could we make the variable "liveReload" and then set it to true by default? Having negations in variable names makes my tiny monkey brain hurt.

ibazel/ibazel.go
```diff
@@ -142,6 +157,7 @@ func (i *IBazel) Cleanup() {
```
achew22: It seems like there should be a step in Cleanup() where the live reload server gets shut down. If not, can you help me understand why if that understanding is wrong?

ibazel/main.go
```diff
@@ -27,6 +27,7 @@ var overrideableBazelFlags []string = []string{
 
 var logToFile = flag.String("log_to_file", "-", "Log iBazel stderr to a file instead of os.Stderr")
 var noLiveReload = flag.Bool("nolive_reload", false, "Disable JavaScript live reload support")
+var profileDev = flag.Bool("profile_dev", false, "Enable development profiling")
```
achew22: Can you make the comment describing this more specifically? For example, one might wonder "How does 'development profiling' differ from the Bazel --profile flag"

What would you think about renaming it to "log_timing_info" (I like green bikesheds)?

ibazel/ibazel.go
gregmagolan: The command line flag was chosen to match the iblaze `nolive_reload` so it would be consistent between the two. Variable here was named to match the flag. Are you thinking change the variable here to `liveReload` and be true by default and keep the flag as `nolive_reload` to be consistent with iBlaze? I like that variable naming better. Agree that negated variables are harder to read.

ibazel/ibazel.go
gregmagolan: The server runs in a goroutine started with `go i.lrserver.ListenAndServe()`. I'm new to go but from what I've read, when the main process exits all goroutines are stopped where they are and resources are cleaned up. This haven't had any issues with the lrserver port not being released on exit either... although iBazel itself sometimes doesn't quit on CTRL-C and I have to kill the process manually on OSX.

Internally, the lrserver is blocking in the http.Server.Serve() call. Could fork the lrserver and add a signal handler to close the http server and then then iBazel could request a close on Cleanup() and wait for the goroutine to end. That being said, it works fine with go handling the lrserver goroutine cleanup as well.

gregmagolan:
Thanks for looking it over @achew22.

The `profile_dev` naming was @alexeagle's preference so that it would be similar but different to the Bazel `profile` flag. This is a work in progress... plan is to have iBazel write build timing results to a user specified file and also open up a port to accept timing results from the browser as well to add to the results. Similar work is being done in parallel in iBlaze.

ibazel/main.go
gregmagolan: The `profile_dev` naming was @alexeagle's preference so that it would be similar but different to the Bazel `profile` flag. This is a work in progress... plan is to have iBazel write build timing results to a user specified file and also open up a port to accept timing results from the browser as well to add to the results. Similar work is being done in parallel in iBlaze.

gregmagolan:
Should add an example of what turning it on looks like in a rule:

```
ts_devserver(
    name = "devserver",
    # Points to a rule that produces a Manifest as its default output
    manifest = ":devsources",
    # This is the path we'll request from the browser, see index.html
    serving_path = "/bundle.js",
    # The devserver can serve our static files too
    static_files = ["index.html", "node_modules/requirejs/require.js"],
    # Turn on live reload
    tags = ["ibazel_live_reload"],
)
```

ibazel/ibazel.go
gregmagolan: So ended up added a Close() function to lrserver on a fork. It was needed for testing as the Bazel instance is spun up and cleaned up all in the test process so can't rely on go to close down the live reload server goroutine.

gregmagolan:
PR updated with basic livereload server test in ibazel_test.go and also added the IBAZEL_LIVERELOAD_URL env variable for the devserver rule to use

bazel-io:

gregmagolan:
@achew22 Would be good to test that the live reload server does not start when the tag is not present. However, not sure what the best way is to update the test file to have different variations of mock query results.

achew22:
I'm working on the ability to run better e2e tests from inside Blaze @ https://github.com/bazelbuild/bazel-integration-testing/pull/7. Lemme get some tests out that use that and you can use them as templates. Does that work for you?

gregmagolan:
@achew22 @alexeagle

So would be good to get this in this week along and make a new iBazel release and get the new devserver in rules_typescript out as well.

Shall I change the `noLiveReload` variable to `liveReload`?

It would be good to get some some version info as well so you could type `ibazel --version` and see what you're running. I'll see if I can put up a PR for that today.

achew22:
Yeah, let's leave the command line flag as --nolive_reload but we can track the internal state as liveReload = false

achew22:
Also, could you rebase your change onto master? A fix went in that should make the travis server work now.

gregmagolan:
Will make a new PR with the requested changes and the profile_dev taken out

PR #71, opened 2017-11-15
Unwatch files no longer returned by query

dougkoch:
We keep track of the files under each watcher, and unwatch any files that are no longer returned by the query associated with that watcher.


dougkoch:
@achew22 Any thoughts on this PR? It has been sitting for a little while.

achew22:
Passed in 2 CIs for Linux. Darwin is not happening tonight.

Issue #45, opened 2017-10-03
avoid version problems for node users


alexeagle:
There are two ways:

1) recommend `PATH=./node_modules/.bin:$PATH` so that if you are in a project directory where ibazel@x.y.z is installed, you get that version. @vsavkin suggests this. But if you're in a child directory (common for bazel users since it does a good job in this case) it doesn't work

2) when ibazel runs, it checks if there is a locally installed copy, and exec()s that one instead. gulp and angular-cli do it this way. @hansl suggests this

ittaiz:
What is the current state?

alexeagle:
Currently you either do a global install, then get the global version anywhere you run the tool, or list ibazel in your package.json#devDependencies, then you have to run it like `./node_modules/.bin/ibazel`

ittaiz:
got it. thanks

mrmeku:
I think this issue is a bit tricky since you can specify multiple targets for bazel to build/run

bazel build :some_target path:some_other_target

Should we only try to pick the local ibazel if they specify one target? Or maybe only if they are in a subdirectory with a local bazel installation (so running ibazel from outside the project directory would do use the global installed version)?

I think we need to think about a really easy to understand heuristic we can use so that we can communicate that to ibazel users

alexeagle:
I don't think it matters what the target patterns are.
You just look in the cwd() where the ibazel is run. If you find a locally installed in `./node_modules` or `../(../)+node_modules` then you use that (following node module resolution semantics). Finally you give up and run the globally installed one.

mrmeku:
So basically we punt on if they have a local installation but invoke ibazel
from a parent directory of their project. I think that's pretty reasonable.

achew22:
Is there a good way to scan for `node_modules` directories other than to recurse up and see if it exists?

alexeagle:
That's what other implementations do, I believe

PR #74, opened 2017-11-26
Search for a local npm installation

dougkoch:
Fixes: #45

### DIFF
--- a/ibazel/main.go
+++ b/ibazel/main.go
@@ -18,9 +18,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


mrmeku:
Shouldn't this logic live within npm/index.js? It seems like it's a npm/node specific concern and having the logic live here wouldn't allow users to manually invoke the bin the want to override this logic

dougkoch:
Good call. I'm not too familiar with npm so it just didn't occur to me to handle this in the package entry point. I'll move the logic there when I get a chance.

dougkoch:
This PR is ready to be reviewed again.

Issue #72, opened 2017-11-16
bazel test fails locally and on TravisCI


gregmagolan:
Not sure where to begin to track this down... failure on Travis looked like this on my PR.

```
INFO: Found 14 targets and 5 test targets...
INFO: Building...
[0 / 1] BazelWorkspaceStatusAction stable-status.txt
[2 / 35] Writing file external/org_golang_x_sys/unix/libgo_default_library.cgo\
_c_lib.lo-2.params
[5 / 55] Compiling external/com_github_google_protobuf/src/google/protobuf/com\
piler/java/java_map_field.cc [for host]
[26 / 74] Compiling external/com_github_google_protobuf/src/google/protobuf/em\
pty.pb.cc [for host]
[57 / 105] Compiling external/com_github_google_protobuf/src/google/protobuf/c\
ompiler/java/java_message_field.cc [for host]
[84 / 132] Compiling external/com_github_google_protobuf/src/google/protobuf/u\
til/internal/utility.cc [for host]
[188 / 243] Still waiting for 2 jobs to complete:
      Running (local):
        SkylarkAction external/go_stdlib_linux_amd64_cgo/bin/go [for host], 10\
 s
      Scheduling:
        Linking external/com_github_google_protobuf/protoc [for host], 10 s
[215 / 291] GoCompile external/io_bazel_rules_go/go/tools/builders/~stripped~a\
sm~/github.com/bazelbuild/rules_go/go/tools/builders/asm.a [for host]
[222 / 291] GoCompile e2e/simple/~stripped~sample~/github.com/bazelbuild/bazel\
-watcher/e2e/sample.a
[244 / 291] Compiling external/org_golang_x_sys/unix/go_default_library.cgo.di\
r/external_org_golang_x_sys_unix_syscall_linux.cgo2.c
PASS: //bazel:go_default_test
PASS: //ibazel/command:go_default_test
PASS: //ibazel:go_default_test
PASS: //e2e:go_default_test
INFO: Elapsed time: 242.429s, Critical Path: 42.66s
java.lang.IllegalStateException
	at com.google.common.base.Preconditions.checkState(Preconditions.java:429)
	at com.google.devtools.build.lib.util.Preconditions.checkState(Preconditions.java:196)
	at com.google.devtools.build.lib.runtime.TestResultAnalyzer.differentialAnalyzeAndReport(TestResultAnalyzer.java:115)
[...]

PR #76, opened 2017-11-27
Bump Travis config to Bazel 0.8.0

dougkoch:
Fixes: #72


alexeagle:
How will ibazel users know they should update?

achew22:
@alexeagle this is the version that we are testing building iBazel with, not testing compatibility with. See #77 for more words on this.

PR #78, opened 2017-11-28
Live reload

gregmagolan:
* Adds off-the-shelf live reload server (github.com/jaschaephraim/lrserver) that is turned on via `ibazel_live_reload` tag in `ts_devserver` run target (requires unreleased rules_typescript devserver changes for flag support).
* Live reload server defaults to port 35729; ports are scanned and first available port between 35729 and 35828 (range of 100) is chosen
* Adds `-nolive_reload` flag which forces live reload server off even if `ibazel_live_reload` is set

Original PR #69

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -17,8 +17,10 @@ package main
 import (
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -26,6 +28,7 @@ import (
 	"github.com/bazelbuild/bazel-watcher/bazel"
 	"github.com/bazelbuild/bazel-watcher/ibazel/command"
 	"github.com/fsnotify/fsnotify"
+	"github.com/gregmagolan/lrserver"
 
 	blaze_query "github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf"
 )
@@ -54,15 +57,17 @@ type IBazel struct {
 
 	debounceDuration time.Duration
 
-	cmd       command.Command
-	args      []string
-	bazelArgs []string
+	cmd        command.Command
+	args       []string
+	bazelArgs  []string
+	liveReload bool
 
 	sigs           chan os.Signal // Signals channel for the current process
 	interruptCount int
 
 	buildFileWatcher  *fsnotify.Watcher
 	sourceFileWatcher *fsnotify.Watcher
+	lrserver          *lrserver.Server
 
 	sourceEventHandler *SourceEventHandler
 
@@ -77,6 +82,7 @@ func New() (*IBazel, error) {
 	}
 
 	i.debounceDuration = 100 * time.Millisecond
+	i.liveReload = true // default to true
 
 	i.sigs = make(chan os.Signal, 1)
 	signal.Notify(i.sigs, syscall.SIGINT, syscall.SIGTERM)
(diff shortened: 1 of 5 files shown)


bazel-io:

dougkoch:
Test this please

dougkoch:
Test this please

ibazel/ibazel.go
```diff
@@ -372,3 +429,15 @@ func (i *IBazel) watchFiles(query string, watcher *fsnotify.Watcher) {
 
 	fmt.Fprintf(os.Stderr, "Watching: %d files\n", successFullWatchFileCount)
 }
+
+func testPort(port uint16) bool {
```
dougkoch: This is nitpicky, but can you fix the indentations in this function?

gregmagolan: Done. Changed to tabs.

gregmagolan:
> Test this please

What tests are you thinking? Added a `TestIBazelRun_livereload` which tests that `IBAZEL_LIVERELOAD_URL` is set correctly and that the live reload port is open and serves up the livereload .js file correctly.

ibazel/ibazel.go
achew22: Can you also run gofmt?

dougkoch:
@gregmagolan "Test this please" isn't a request for you, it's the phrase I have to type to kick off a new Jenkins build. The Travis builds happen automatically, but the Jenkins builds don't. I'm happy with your PR and I'm about to merge it.

ittaiz:
Greg,
I'm pretty sure Doug was triggering the CI bot on your commit

ibazel/ibazel_test.go
```diff
@@ -283,6 +290,41 @@ func TestIBazelRun_notifyPreexistiingJobWhenStarting(t *testing.T) {
...
+	i, err := New()
+	if err != nil {
+		t.Errorf("Error creating IBazel: %s", err)
+	}
+	defer i.Cleanup()
+
+	i.run("//path/to:target")
+
+	livereloadUrl := os.Getenv("IBAZEL_LIVERELOAD_URL")
+	var validUrl = regexp.MustCompile(`^http\:\/\/localhost\:[0-9]+\/livereload\.js\?snipver\=1$`)
```
achew22: I would prefer you use the more idiomatic golang

`validUrl := regexp.MustCompile("^http\:\/\/localhost\:[0-9]+\/livereload\.js\?snipver\=1$")`

2 changes in there.

1) Using `:=` to define and set a variable in one pass
2) Using double quotes for strings. backtick is great but it is easy to accidentally turn into a multi-line string.

gregmagolan: K. Switched to :=. Thanks.
Trying double quotes breaks the compile however with the error ` ibazel/ibazel_test.go:303:40: unknown escape sequence`. Might be that backticks are required for regexp. https://golang.org/pkg/regexp/. Not sure as I'm new to Go.

gregmagolan:
@achew22 Ran gomt as well

ibazel/ibazel_test.go
dougkoch: You can double escape and it will work: `validUrl := regexp.MustCompile("^http\\:\\/\\/localhost\\:[0-9]+\\/livereload\\.js\\?snipver\\=1$")`

gregmagolan: That works. Thanks.

dougkoch:
Test this please

dougkoch:
@gregmagolan Thanks for the changes, just waiting on Travis CI before we can merge this. It has a very long OS X backlog because of their limited Mac hardware (see https://www.traviscistatus.com/)

PR #80, opened 2017-11-30
Sync package.json contributors with /CONTRIBUTORS

gregmagolan:
A bit annoying that these both need to be maintained but may as well keep them in sync.


bazel-io:

gregmagolan:
Whoops. Messed that up.

achew22:
Greg, thanks for putting this in. I'm actually writing something right now that will make this kind of change unnecessary. Can you give me an hour or two and tell me what you think?

PR #82, opened 2017-12-02
Move e2e tests to bazel-integartion-testing.

achew22:
Tests now run through full versions of Bazel. This will allow testing
against a matrix of Bazel versions while only maintaining build
compatibility with a single Bazel and rules_go.


dougkoch:
LGTM, but I'll defer to Alex or Dan to approve this.

Issue #49, opened 2017-10-09
broken when running from WORKSPACE/subdirectory


alexeagle:
I get `lstat` errors about no such file or directory, it seems like the paths being watched are still workspace-relative but then don't exist relative to the pwd when in a subdirectory of the workspace.

PR #83, opened 2017-12-03
Watch absolute paths

dougkoch:
Fixes: #49

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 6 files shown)


ibazel/ibazel.go
```diff
@@ -99,6 +106,35 @@ func New() (*IBazel, error) {
...
+			path = volume
+		}
+
+		// Check if we're at the workspace path
+		if _, err := os.Stat(filepath.Join(path, "WORKSPACE")); !os.IsNotExist(err) {
+			return path, nil
+		}
+
+		// If we've reached the root, then we know the cwd isn't within a workspace
+		if path == volume {
```
achew22: This should fail when called from inside a test. If it doesn't then it is depending on a fragile implementation detail of Bazel's test launcher that will change in the future and break.

achew22:
#89 should also function as a template for writing an e2e test. Take the simple test, cd into a subdir, start ibazel, cd back and then run the rest of the test.

ibazel/ibazel.go
dougkoch: Are you implying that this should just default to the cwd if a WORKSPACE isn't found (and let bazel give the error)? Or would you suggest some other way to keep this from failing in a test?

achew22: I think this is an interface, "WORKSPACEFinder", or something and that there can be a trivial testing fake that returns $CWD or empty string or something as the path.

PR #87, opened 2017-12-05
WIP: Profiler

gregmagolan:


bazel-io:

gregmagolan:
@alexeagle Here's the start of the iBazel profiler. Next step is to start receiving data from the browser I think. Who's a good person to look over my Go code for style and correctness?

achew22:
@gregmagolan/@alexeagle, can you help me understand the additional value of adding profiling to iBazel? Isn't most of the time going to be in Bazel and that already has a profiler?

gregmagolan:
> @gregmagolan/@alexeagle, can you help me understand the additional value of adding profiling to iBazel? Isn't most of the time going to be in Bazel and that already has a profiler?

Working towards getting development RTT timing data back from the browser to put into the profile file so you can track over time your full build time as well as developer RTT from file change to interactive browser.

gregmagolan:
It would be great to have a badge for GitHub with the build time and RTT on file change as well as a history so you can catch regressions in build performance.

alexeagle:
@achew22 sorry I think we missed reviewing the design with you.

Bazel's `--profile` option does collect useful data to understand the critical path of a single build.

Like Greg says, we want to measure round-trip time of any development workflow, which means measuring from the time the user touches the file, to when they can operate the app, and measuring a few iterations of this to smooth out noise. This means we need some way for the program being run to contribute timing events of its own.

Yi Qi is implementing the same thing in iblaze, so our plan is to keep the architecture of the two the same.

achew22:
@gregmagolan, I have some changes that are going to land probably tonight that will make putting this code in a lot easier. Would you mind holding off till at least tomorrow?

@alexeagle, no worries. You may want to inform me of your master plan next time before it starts getting implemented :wink:

gregmagolan:
@achew22 No prob. It’s going to be a few more days before this is ready. What do you have coming?

achew22:
@gregmagolan, I just created the PR as #92. The realization I had is that most of the desired functionality from here on out is pretty modular revolving around the state machine + a few additional things. The profiler you would like to add is a great addition and I think fits really nicely into the lifecycle concept (please tell me if I'm wrong).

Additionally I added an end to end test to verify the livereload server is behaving properly in that PR. I don't know how I was going to break that in the future, but I assure you it will happen. Hopefully it shouldn't do that now.

gregmagolan:
Will open a new PR with the profile as a lifecycle hook

PR #88, opened 2017-12-05
Don't spam the travis log with bazel progress

alexeagle:


.test-bazelrc
```diff
@@ -2,3 +2,6 @@
 # They are used by Travis CI and by non-Bazel test scripts.
 build --verbose_failures --sandbox_debug --test_output=errors --spawn_strategy=standalone --genrule_strategy=standalone
 test --test_strategy=standalone
+
+# Don't spam the travis log
+build --noshow_progress
```
achew22: Could you make this

`build:ci --noprogress`

then modify .travis.yaml to include a `--config=ci` flag?

achew22: Oh, geez. NM. I want to do this a different way. I'll send you a PR in a sec

Issue #18, opened 2017-07-28
Automatically apply buildozer commands from warnings


ittaiz:
Hi,
I've heard that the internal version of Bazel-watcher (iBazel) knows to watch for log warnings in Bazel outputs which are buildozer commands and to automatically apply them.
The need for this (for me) is to be able to automatically apply strict java/scala deps warnings.
This means a developer can use code from transitive targets, the rule will warn against it with a buildozer command (saying which target needs to be added explicitly where) and ibazel will automatically apply it.
Is there an interest to extend ibazel to that area?

achew22:
Huh, this is a very interesting idea. I don't think I've yet run into that in iBazel but it doesn't mean that it doesn't do it. Do you have any ideas on how this would work? Are there any rules_ that are emitting buildozer commands to execute in case of failure?

In the specific case of Gazelle it might be reasonable to just `blaze run //:gazelle` whenever there is an import error but I'm not sure how to actually detect that. WDYT?

ittaiz:
rules_scala does [this](https://github.com/bazelbuild/rules_scala/blob/master/third_party/plugin/src/main/io/bazel/rulesscala/dependencyanalyzer/DependencyAnalyzer.scala#L66)
What is Gazelle?
BTW, @cgrushko is the one who told me about this feature in internal ibazel so maybe he has more info

cgrushko:
IIUC it's nothing more than ibazel looking at the output of Bazel and running some commands when it sees something it recognizes (i.e., a regex --> command map).

mprobst:
The internal tool parses error messages from the compilation job for well
known patterns and then runs the suggested fixer tool (if there is any).
For Go code indeed, you'd probably just run gazelle, but for Java code you
often need very specific fixes (add a dependency to this target or that).

ittaiz:
Given this PR <https://github.com/bazelbuild/bazel/pull/3677> which adds
printing the buildozer command on strict_java_deps warn/error.
Can ibazel parse the message and just run the buildozer command?

achew22:
That seems like an extremely reasonable thing to do. If you wanted to implement it with a whitelist of regexps (to prevent running something scary) scanning the output from Bazel I think you could implement that without a huge amount of effort.

Why don't you take a whack at it? If you run into problems, feel free to comment on here.

achew22:
Useful notes to future implementers:

[...]

PR #92, opened 2017-12-08
Add the concept of lifecycle hooks

achew22:
Also migrate livereload into being the first consumer of these lifecycle
hooks. This should allow the core of iBazel to be relatively simple but
additional modules be made available to the end user in any boolean
combination.

I'm open and supportive of adding more hooks into the lifecycle of an
iBazel session. If there is something you would like to see hooked, add
it to the Lifecycle interface with a before/after version.

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -17,18 +17,16 @@ package main
 import (
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"os/signal"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/bazelbuild/bazel-watcher/bazel"
 	"github.com/bazelbuild/bazel-watcher/ibazel/command"
+	"github.com/bazelbuild/bazel-watcher/ibazel/live_reload"
 	"github.com/fsnotify/fsnotify"
-	"github.com/gregmagolan/lrserver"
 
 	blaze_query "github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf"
 )
@@ -57,21 +55,20 @@ type IBazel struct {
 
 	debounceDuration time.Duration
 
-	cmd        command.Command
-	args       []string
-	bazelArgs  []string
-	liveReload bool
+	cmd       command.Command
+	args      []string
+	bazelArgs []string
 
 	sigs           chan os.Signal // Signals channel for the current process
 	interruptCount int
 
 	buildFileWatcher  *fsnotify.Watcher
 	sourceFileWatcher *fsnotify.Watcher
-	lrserver          *lrserver.Server
 
 	filesWatched map[*fsnotify.Watcher]map[string]bool // Inner map is a surrogate for a set
 
 	sourceEventHandler *SourceEventHandler
+	lifecycleListeners []Lifecycle
 
 	state State
 }
@@ -85,10 +82,17 @@ func New() (*IBazel, error) {
 
 	i.debounceDuration = 100 * time.Millisecond
 	i.filesWatched = map[*fsnotify.Watcher]map[string]bool{}
-	i.liveReload = true // default to true
 
 	i.sigs = make(chan os.Signal, 1)
 	signal.Notify(i.sigs, syscall.SIGINT, syscall.SIGTERM)
+
+	i.lifecycleListeners = []Lifecycle{
+		live_reload.New(),
+	}
+
+	for _, l := range i.lifecycleListeners {
+		l.Setup()
+	}
 
 	go func() {
 		for {
@@ -140,10 +144,6 @@ func (i *IBazel) SetBazelArgs(args []string) {
 	i.bazelArgs = args
 }
 
-func (i *IBazel) SetLiveReload(liveReload bool) {
-	i.liveReload = liveReload
-}
-
 func (i *IBazel) SetDebounceDuration(debounceDuration time.Duration) {
 	i.debounceDuration = debounceDuration
 }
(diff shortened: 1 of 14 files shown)


gregmagolan:
@achew22 I like it. The profiler in #87 could be wired up to this concept with some modifications. The complications that I'd have with this as is are:

1) I'm creating a profile event if and when the livereload is triggered which would require lifecycle events to be able to create new lifecycle events.

2) Change detection events pass the name of the change which wouldn't fit in the beforeEvent() afterEvent() pattern.

I think a few changes could make this more flexible. With different lifecycle event types requiring different parameters and the interface being the natural place to define those parameters, the event types and parameters could be defined by the lifecycle interface as such:

```
type Lifecycle interface {
	// Initialize is called when iBazel starts
	Initialize()

	// RunTarget is called when a target is being setup to run
	RunTarget(rule *blaze_query.Rule)

	// ChangeDetected is called when a change is detected
	// changeType: "source"|"graph"
	// name: name of source file or build graph change
	ChangeDetected(changeType string, name string)

	// Called before/after running a command
	// commandType: "build"|"test"|"run"
	// success: true when command succeeded, false when failed
	BeforeCommand(commandType string)
	AfterCommand(commandType string, success bool)

	// Called after a livereload is triggered
	ReloadTriggered()

	// Cleanup is called when iBazel is being shutdown
	Cleanup()
```

For triggering a ReloadTriggered() event from within an AfterCommand() event, a Lifecycle interface could potentially be passed into to each event which could then be used to create other events as such:

`AfterCommand(lifecycle Lifecycle, commandType string, success bool)`

Thoughts?

ittaiz:
Andres,
With respect to #18 (buildozer auto apply) do you think it should be a hook
or is this not a relevant use case for these hooks?

ibazel/live_reload/server.go
```diff
@@ -0,0 +1,108 @@
...
+				}
+			}()
+			url := fmt.Sprintf("http://localhost:%d/livereload.js?snipver=1", port)
+			os.Setenv("IBAZEL_LIVERELOAD_URL", url)
+			return
+		}
+	}
+	fmt.Fprintf(os.Stderr, "Could not find open port for live reload server\n")
+}
+func (l *LiveReloadServer) Cleanup() {}
```
mrmeku: Can you leave a comment as to why its not necessary to kill the reload server and unset the IBAZEL_LIVERELOAD_URL within the cleanup step? I didn't expect this method to be empty

ibazel/lifecycle.go
```diff
@@ -0,0 +1,25 @@
...
+	// used. You can call methods on the ibazel object in this context to set
+	// additional environment variables to be passed into the client or to
+	// retrigger action.
+	Setup()
+	// Cleanup is your opportunity to clean up open sockets or connections.
+	Cleanup()
+
+	// Before running an "event" where name = (build|test|run).
+	BeforeEvent(string)
+	AfterEvent(string)
```
mrmeku: Seems like you should pass a bit more context. For example, if the live reload server wanted to make an optimization by rereading only changed files, it would need to know which build targets were affected by the build event. Or if I'm missing something maybe you could add a comment to explain how to hook into metadata like that

ibazel/ibazel.go
```diff
@@ -140,19 +144,31 @@ func (i *IBazel) SetBazelArgs(args []string) {
...
-	if i.lrserver != nil {
-		i.lrserver.Close()
+	for _, l := range i.lifecycleListeners {
+		l.Cleanup()
+	}
+}
+
+func (i *IBazel) targetDecider(rule *blaze_query.Rule) {
+	for _, l := range i.lifecycleListeners {
+		l.TargetDecider(rule)
```
mrmeku: Should target decider return a bool of whether it wants to receive events? Otherwise each lifecycle listener will need to maintain state and have logic to perform a noop on BeforeEvent and AfterEvent calls

achew22:
@gregmagolan 

> I'm creating a profile event if and when the livereload is triggered which would require lifecycle events to be able to create new lifecycle events.

Live reload is currently triggered by an event. Could we have have the profiler note the time when they received the before/after event and say "that's how long it took"? The profiler may be a special case where we add some profiling code into one of the lifecycle listeners, but I'd like to avoid it if possible.

> Change detection events pass the name of the change which wouldn't fit in the beforeEvent() afterEvent() pattern.

Thanks for the feedback! I'll be updating my commit in just a second to reflect your desired changes. One of the nice things about this interface is that all the implementors are going to be in the same repo so if we want to make changes we still can. In addition to the success information I would like to pass the action's run output. It might be possible to do some nifty things with the test output where you print out a reformatted form of the test output that's easier to read. Or, in the case of a build failure, the liveserver pushes a message to the browser that says "Hey, you had a typo on line 12 of `foo.ts` and your build didn't work".

@ittaiz,  that's one of the many things I think we can solve with this. We can write a relatively simple post execution hook that will get the `blaze $COMMAND` output buffer after the program exits, parse it for magic strings, and then invoke `bazel run //:gazelle` or `buildozer add ...` or whatever it is you think should be run afterward.

@mrmeku, thanks for the review. Really good comments which I will address individually when I upload more code in a little bit.

ittaiz:
Awesome!
I'll try to take it as a motivation for picking the buildozer feature after
this is merged

ibazel/ibazel.go
achew22: Yes, that seems more correct than this. I think that might get a little complicated tracking that in a generic form so I'm going to return to it a little later. Todo added with context of why I don't wanna do it right now.

ibazel/lifecycle.go
achew22: In the particular case of the live reload server, there can only be one target being run at a time. As a result, in the particular case of live reload the answer to "what target/files should I reload" is always going to be all of them.

There may, however. be some special logic that is useful in a `ibazel test` scenario. Lemme think on that.

ibazel/live_reload/server.go
achew22: Good eyes. I messed that one up real good!

PR #103, opened 2017-12-17
Move e2e output assertions into helper

achew22:
This allows much longer timeouts to be used while still achieving faster
runtimes. The way I do this is by moving the assertion from the test
intointo the e2e helper. This way the call to get the output knows what
it is waiting for. Since it knows what it is waiting for, if it gets it
before time is up then it's free to continue onto the next step. If it
doesn't get it then it will wait until the timeout arrives.

### DIFF
--- a/e2e/ibazel.go
+++ b/e2e/ibazel.go
@@ -7,22 +7,32 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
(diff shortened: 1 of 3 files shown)


achew22:
@dougkoch, this should help make things less flakey if only by virtue of increasing the timeout on actions. PTAL

PR #112, opened 2018-01-09
Optional config flag to be passed to bazel (for CI use with --config=ci)

gregmagolan:
Allow for passing `--config=ci` to Bazel for running ibazel within CI

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -28,6 +28,7 @@ import (
 )
 
 var bazelPath = flag.String("bazel_path", "bazel", "Path to the bazel binary to use for actions")
+var config = flag.String("config", "", "Config flag passed to bazel")
 
 type Bazel interface {
 	SetArguments([]string)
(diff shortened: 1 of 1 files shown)


bazel-io:

gregmagolan:
@alexeagle This is needed for running the benchmark runner in CI

achew22:
I think you could also add it to this list and it would have a pretty similar outcome https://github.com/gregmagolan/bazel-watcher/blob/ea76f5aed3d41c360889c6d992426c3fe05f1744/ibazel/main.go#L28

gregmagolan:
> I think you could also add it to this list and it would have a pretty similar outcome https://github.com/gregmagolan/bazel-watcher/blob/ea76f5aed3d41c360889c6d992426c3fe05f1744/ibazel/main.go#L28

Ok. Sweet. Will make another PR. Thanks.

PR #113, opened 2018-01-09
Optional config flag to be passed to bazel (for CI use with --config=ci)

gregmagolan:


bazel-io:

ibazel/main.go
```diff
@@ -26,6 +26,7 @@ var Version = "Development"
 
 var overrideableBazelFlags []string = []string{
 	"--test_output=",
+	"--config-",
```
achew22: This should probably be `--config=`

gregmagolan:
> Can you help me understand the circumstances where you need this?

When running bazel within a CI environment such as circle CI, the `--config=ci` flag is needed. @alexeagle knows more of the details of why but I believe it changes how bazel calculates the available memory for a dockerized environment.

Thanks for catching my typo.

gregmagolan:
ibazel-benchmark-runner runs ibazel within CI so it will pass `--config=ci` to iBazel

alexeagle:
I would expect the opposite: ibazel could block me setting certain bazel flags that it controls, but should allow me to pass anything else I want. What's the motivation for using a whitelist?

achew22:
This is exactly this list is, a whitelist of flags that are parsed beyond the normal "stuff after `--` is passed down the line"

alexeagle:
I'm curious what's the motivation for using a whitelist instead of a
blacklist?  I think in this design, we'll have a lot more options to add to
this list.

achew22:
What do you blacklist arguments wise? This is a list of arguments who, when they appear anywhere in the command line invocation, are passed into Bazel. Bazel has on the order of 400 flags so I didn't go and enumerate them but that is probably where we will end up on this.

gregmagolan:
@achew22 Can you merge this in and make a 0.3.1 release on npm? Would be good to have the benchmark runner script depend on a proper released version of iBazel.

We can discuss the issue of whitelist vs. blacklist elsewhere.

achew22:
Test this please

gregmagolan:
@achew22 Ping for 0.3.2?

mprobst:
iblaze has a long list of all known flags of blaze, including whether or not they take a value, and accepts and passes all of those on.

Issue #52, opened 2017-10-17
data race in ibazel


jmhodges:
Here's the stacktrace of a data race with the latest ibazel e8cbba37f5aa56f36e8561ec64128008c1f2cbe1. It came up in normal use.

It's the same one I mentioned in https://github.com/bazelbuild/bazel-watcher/pull/35

```
==================
WARNING: DATA RACE
Read at 0x00c42008c788 by goroutine 11:
  github.com/bazelbuild/bazel-watcher/ibazel/command.(*defaultCommand).IsSubprocessRunning()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/command/command.go:75 +0xae
  main.(*IBazel).handleSignals()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/ibazel.go:96 +0x478
  main.New.func1()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/ibazel.go:83 +0x38

Previous write at 0x00c42008c788 by main goroutine:
  os/exec.Command()
      /usr/local/Cellar/go/1.9.1/libexec/src/os/exec/exec.go:143 +0x157
  github.com/bazelbuild/bazel-watcher/ibazel/command.start()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/command/command.go:58 +0x397
  github.com/bazelbuild/bazel-watcher/ibazel/command.(*defaultCommand).Start()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/command/default_command.go:64 +0x161
  github.com/bazelbuild/bazel-watcher/ibazel/command.(*defaultCommand).NotifyOfChanges()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/command/default_command.go:75 +0x46
  main.(*IBazel).run()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/ibazel.go:281 +0x155
  main.(*IBazel).(main.run)-fm()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/ibazel.go:158 +0x5f
  main.(*IBazel).iteration()
      /Users/jmhodges/src/github.com/bazelbuild/bazel-watcher/ibazel/ibazel.go:218 +0x656
  main.(*IBazel).loop()
[...]

PR #121, opened 2018-02-10
run the race detector in CI

jmhodges:
Demonstrates the bug in #52


bazel-io:

achew22:
Test this please

jmhodges:
Not sure how to adjust the ci.bazel.build CI configs to include `--features=race`, so that'll need it's own work, probably.

jmhodges:
I rebased this against master and now the Travis CI output isn't great and elides why the tests are failing.

It's unclear why.

jmhodges:
Added some better flags to improve the Travis CI logs.

It looks like the build uses `git describe` when it maybe should be using `git rev-parse` https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything

jmhodges:
(I suspect this happens when anyone makes a PR from a fork that is not the bazelbuild fork because of the vagaries of Travis CI.)

jmhodges:
Yeah, tools/workspace_status.sh can't work if someone other than bazelbuild makes the PR. Adding `git fetch --tags` in there if it's run in Travis CI doesn't fix it because the fork the PR is from doesn't have the tags.

So, the Travis CI build is broken for anyone trying to contribute to bazelbuild from outside and its unclear how it should be fixed as a contributor. 

It's also still unclear how to get the ci.bazelbuild configs changed.

jmhodges:
I hope, at the very least, the maintainers can see what changes in their own local systems (adding `--features=race` to their `bazel test` runs) will demonstrate the bug in #52.

jmhodges:
Okay, I hacked around the workspace_status.sh problem to get this PR to fail on the race conditions again.

Happy hunting.

achew22:
@jmhodges, I was just reading through the `rules_go` source. It looks like we can set `race = "on"` in the `bazel_go_integration_test` target and it fails in the same way. That might be a good way to go about getting it running in race mode in both ci.bazel.build and travis

jmhodges:
Your call.

I'm suspicious of target specific settings because invariably another target gets written and people forget to add it there or can't add there because it needs to be off in other contexts or there's just no easy way to do so.  I don't know how ci.bazelbuild's configs are controlled but I imagine (given my limited) understanding of how Goog handles these things) they are checked in somewhere and it would be relatively easy to toss in the flag over there?

I think I'm not going to edit this patch right now since it can't be merged without work elsewhere and has successfully reproduced the original problem it was meant to demonstrate.

jmhodges:
(Plus I imagine there are unit tests that would also trigger races and are, generally, easier to debug in that case)

PR #131, opened 2018-07-30
Output Runner and Multiple Runs

borkaehw:
# Output Runner
## Issue
```
```
or
```
```
The build fails and hangs.
## Fix
```
[y/N]
```

To enable output runner, use flag `-run_output=config_file_name.json`. The config file format is provided in `ibazel/output_runner/output_runner_test.json`.

# Multiple Runs
## Issue
iBazel currently only supports `run` for single target. Running multiple targets requires separate Bazel command executions and busy waiting for build processes.
## Fix
Implement a `mrun` command to run multiple targets with iBazel
```
ibazel mrun target1 --arg=t1_arg1 --arg=t1_arg2 target2 --arg=t2_arg1 --arg=t2_arg2
```
Append arbitrary number of targets after `mrun` with the corresponding arguments, prefixed with `--arg=`.


borkaehw:
I signed it!

achew22:
Hey @borkaehw, This looks like a really great patch to contribute! Thanks for sending what you have in. If you rebase your change off master it might be a bit easier to take a look at.

borkaehw:
New PR #133 is created. Cleaning up the mess.

PR #133, opened 2018-07-31
Output Runner and Multiple Runs

borkaehw:
# Output Runner
## Issue
```
```
or
```
```
The build fails and hangs.
## Fix
```
[y/N]
```

To enable output runner, use flag `-run_output=config_file_name.json`. The config file format is provided in `ibazel/output_runner/output_runner_test.json`.

# Multiple Runs
## Issue
iBazel currently only supports `run` for single target. Running multiple targets requires separate Bazel command executions and busy waiting for build processes.
## Fix
Implement a `mrun` command to run multiple targets with iBazel
```
ibazel mrun target1 --arg=t1_arg1 --arg=t1_arg2 target2 --arg=t2_arg1 --arg=t2_arg2
```
Append arbitrary number of targets after `mrun` with the corresponding arguments, prefixed with `--arg=`.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -15,10 +15,12 @@
 package bazel
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
@@ -35,9 +37,9 @@ type Bazel interface {
 	WriteToStdout(v bool)
 	Info() (map[string]string, error)
 	Query(args ...string) (*blaze_query.QueryResult, error)
-	Build(args ...string) error
-	Test(args ...string) error
-	Run(args ...string) (*exec.Cmd, error)
+	Build(args ...string) (*bytes.Buffer, error)
+	Test(args ...string) (*bytes.Buffer, error)
+	Run(args ...string) (*exec.Cmd, *bytes.Buffer, error)
 	Wait() error
 	Cancel()
 }
@@ -69,17 +71,38 @@ func (b *bazel) WriteToStdout(v bool) {
 	b.writeToStdout = v
 }
 
-func (b *bazel) newCommand(command string, args ...string) {
+func (b *bazel) newCommand(command string, args ...string) (*bytes.Buffer, *bytes.Buffer) {
 	b.ctx, b.cancel = context.WithCancel(context.Background())
 
 	args = append([]string{command}, args...)
+
+	if b.writeToStderr || b.writeToStdout {
+		containsColor := false
+		for _, arg := range args {
+			if strings.HasPrefix(arg, "--color") {
+				containsColor = true
+			}
+		}
+		if !containsColor {
+			args = append(args, "--color=yes")
+		}
+	}
 	b.cmd = exec.CommandContext(b.ctx, *bazelPath, args...)
+
+	stdoutBuffer := new(bytes.Buffer)
+	stderrBuffer := new(bytes.Buffer)
 	if b.writeToStderr {
-		b.cmd.Stderr = os.Stderr
+		b.cmd.Stderr = io.MultiWriter(os.Stderr, stdoutBuffer)
+	} else {
+		b.cmd.Stderr = stdoutBuffer
 	}
 	if b.writeToStdout {
-		b.cmd.Stdout = os.Stdout
+		b.cmd.Stdout = io.MultiWriter(os.Stdout, stderrBuffer)
+	} else {
+		b.cmd.Stdout = stderrBuffer
 	}
+
+	return stdoutBuffer, stderrBuffer
 }
 
 // Displays information about the state of the bazel process in the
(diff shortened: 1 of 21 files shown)


borkaehw:
Hey @achew22, sorry for the mess yesterday. I created another PR, hopefully this is clearer. I am happy to take any suggestion for improvement. Thanks in advance.

alexeagle:
What's the tradeoff of just changing the behavior of `run` command? Is it that we can't distinguish a second target to run from an argument to the first program? I'm not sure that your scheme for disambiguating args from additional targets-to-run is sufficient either.
I don't have a better proposal after thinking for a couple of minutes - you could read multiple lines from stdin but I think there may be other reasons to pass things to stdin of ibazel? and might be too subtle from the command line, e.g. `echo -n "//path/to:target1 --arg1 --arg2\n//path/to:target2 --arg3 --arg4" | ibazel run`

Could you clarify who is expected to use `-run_output=config_file_name.json` - is that any time a user wants to run ibazel with the fixer enabled?

bazel/bazel.go
```diff
@@ -69,17 +71,38 @@ func (b *bazel) WriteToStdout(v bool) {
...
 }
 
-func (b *bazel) newCommand(command string, args ...string) {
+func (b *bazel) newCommand(command string, args ...string) (*bytes.Buffer, *bytes.Buffer) {
 	b.ctx, b.cancel = context.WithCancel(context.Background())
 
 	args = append([]string{command}, args...)
+
+	if b.writeToStderr || b.writeToStdout {
+		containsColor := false
```
alexeagle: how is this color change related to the rest of the PR? could it be separated out?

ibazel/ibazel.go
```diff
@@ -37,9 +42,11 @@ var osExit = os.Exit
 var bazelNew = bazel.New
 var commandDefaultCommand = command.DefaultCommand
 var commandNotifyCommand = command.NotifyCommand
+var mrunToFiles = flag.Bool("mrunToFiles", false, "Log mrun to file for simpler log reading")
```
alexeagle: the need for this feature makes me wonder, if you're going to open two or more log files to see the output, why not just run different `ibazel` commands in one or more terminals?

I think the answer is that it's hard to script around "start these three programs in three terminals" in any portable way...

ibazel/ibazel.go
```diff
@@ -229,6 +266,13 @@ func (i *IBazel) Run(target string, args []string) error {
 	return i.loop("run", i.run, []string{target})
 }
 
+// Run the specified target (singular) in the IBazel loop.
+func (i *IBazel) RunMulitple(args, target []string, debugArgs [][]string) error {
```
alexeagle: spelling: Multiple

ibazel/ibazel.go
```diff
@@ -300,38 +353,112 @@ func (i *IBazel) iteration(command string, commandToRun runnableCommand, targets
...
 		i.beforeCommand(targets, command)
-		err := commandToRun(targets...)
-		i.afterCommand(targets, command, err == nil)
+		outputBuffer, err := commandToRun(targets...)
+		i.afterCommand(targets, command, err == nil, outputBuffer)
+		i.state = WAIT
+	}
+}
+
+func (i *IBazel) iterationMultiple(command string, commandToRun runnableCommands, targets []string, debugArgs [][]string, argsLength int) {
```
alexeagle: wow a ton of code is added to this file - is it possible to simplify?

bazel/bazel.go
jjudd: It isn't. I'm pretty sure this is a leftover remnant from an internal modification to make sure ibazel always used color when running bazel commands. It can be pulled out.

jjudd:
Pulling in @BryantN for the mrun vs run change. We were split when we made the change and he was the main proponent for mrun.

ibazel/ibazel.go
borkaehw: I think we are providing this option for those who want to see the outputs, while others who don't need the outputs can use just one terminal for `ibazel`. The other reason is that if you're running ten targets but only need two log files, you can just have two terminals instead of ten.

ibazel/ibazel.go
borkaehw: It's possible. Since `test`, `build`, and `run` share a few functions, `mrun` has its own functions with different signatures but do similar things, we can try to merge those functions and make `mrun` a special case. It requires a lot of works and involves changing code for `test` and `build`. Should we do it?

borkaehw:
@alexeagle Yes. `-run_output=config_file_name.json` is a file that users can define what `bazel` outputs they are watching for. The only use case for us so far is to watch for `buildozer` command and fix the deps problem, we might encounter other use case.

alexeagle:
Hey @borkaehw - what happened with the `multi_binary` proposal to capture the `mrun` use case? Do you have any write-up, or should I start one?

Andrew gathered some notes
```
[multi_binary(
	name = “all_my_binaries”,
	binaries = {
		“//path/to/binary1”: {
			“args”: mode + [“a”, “b”, “c”],
		},
		“//path/to/binary2”, {
			“args”: mode + [“1”, “2”, “3”],
		},
	},
) for mode in [[], [“--debug”]]]

multi_test(
	name = “tests”,
	binaries = “:all_my_binaries”
	test = “:test”,	
)
```

borkaehw:
Hey @alexeagle, thanks for asking. We are currently tackling this Bazel issue
https://github.com/bazelbuild/bazel/issues/2832
Once we have the multiplex worker ready, multiple run seems to be a great project.

Feel free to start one though.

PR #134, opened 2018-08-02
Output Runner

borkaehw:
## Issue
```
```
or
```
```
The build fails and hangs.
## Fix
```
[y/N]
```

To enable output runner, use flag `-run_output=config_file_name.json`. The config file format is provided in `ibazel/output_runner/output_runner_test.json` as an example.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -15,10 +15,12 @@
 package bazel
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
@@ -35,9 +37,9 @@ type Bazel interface {
 	WriteToStdout(v bool)
 	Info() (map[string]string, error)
 	Query(args ...string) (*blaze_query.QueryResult, error)
-	Build(args ...string) error
-	Test(args ...string) error
-	Run(args ...string) (*exec.Cmd, error)
+	Build(args ...string) (*bytes.Buffer, error)
+	Test(args ...string) (*bytes.Buffer, error)
+	Run(args ...string) (*exec.Cmd, *bytes.Buffer, error)
 	Wait() error
 	Cancel()
 }
@@ -69,17 +71,38 @@ func (b *bazel) WriteToStdout(v bool) {
 	b.writeToStdout = v
 }
 
-func (b *bazel) newCommand(command string, args ...string) {
+func (b *bazel) newCommand(command string, args ...string) (*bytes.Buffer, *bytes.Buffer) {
 	b.ctx, b.cancel = context.WithCancel(context.Background())
 
 	args = append([]string{command}, args...)
+
+	if b.writeToStderr || b.writeToStdout {
+		containsColor := false
+		for _, arg := range args {
+			if strings.HasPrefix(arg, "--color") {
+				containsColor = true
+			}
+		}
+		if !containsColor {
+			args = append(args, "--color=yes")
+		}
+	}
 	b.cmd = exec.CommandContext(b.ctx, *bazelPath, args...)
+
+	stdoutBuffer := new(bytes.Buffer)
+	stderrBuffer := new(bytes.Buffer)
 	if b.writeToStderr {
-		b.cmd.Stderr = os.Stderr
+		b.cmd.Stderr = io.MultiWriter(os.Stderr, stdoutBuffer)
+	} else {
+		b.cmd.Stderr = stdoutBuffer
 	}
 	if b.writeToStdout {
-		b.cmd.Stdout = os.Stdout
+		b.cmd.Stdout = io.MultiWriter(os.Stdout, stderrBuffer)
+	} else {
+		b.cmd.Stdout = stderrBuffer
 	}
+
+	return stdoutBuffer, stderrBuffer
 }
 
 // Displays information about the state of the bazel process in the
@@ -99,13 +122,13 @@ func (b *bazel) newCommand(command string, args ...string) {
 func (b *bazel) Info() (map[string]string, error) {
 	b.WriteToStderr(false)
 	b.WriteToStdout(false)
-	b.newCommand("info")
+	stdoutBuffer, _ := b.newCommand("info")
 
-	info, err := b.cmd.Output()
+	err := b.cmd.Run()
 	if err != nil {
 		return nil, err
 	}
-	return b.processInfo(string(info))
+	return b.processInfo(stdoutBuffer.String())
 }
 
 func (b *bazel) processInfo(info string) (map[string]string, error) {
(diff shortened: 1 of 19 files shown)


bazel/bazel.go
```diff
@@ -69,17 +71,38 @@ func (b *bazel) WriteToStdout(v bool) {
 	b.writeToStdout = v
 }
 
-func (b *bazel) newCommand(command string, args ...string) {
+func (b *bazel) newCommand(command string, args ...string) (*bytes.Buffer, *bytes.Buffer) {
 	b.ctx, b.cancel = context.WithCancel(context.Background())
 
 	args = append([]string{command}, args...)
+
+	if b.writeToStderr || b.writeToStdout {
```
borkaehw: Since we are using output buffer for `stdout` and `stderr`, we lose color if we don't explicitly specify `--color`.

alexeagle:
I think it's not sufficiently ergonomic to ask users to always pass the path to the json file everytime they run ibazel.
Maybe there could be a default location(s) we look for such a file, for example `.bazel_fix_commands.json` in the workspace root or something.
Alternately, maybe if no json file is provided, we could have some default regex that would match things like

```
Blah blah error.
To fix, run:
$ some_fixer_command --argument
```

Then you don't need a JSON file if your tool outputs something that matches the default regex(es)

ittaiz:
+1 to both suggestions

jjudd:
I like both suggestions. Any reason not to implement both? 

Also, how do we feel about `^(buildozer) '(.*)'\\s+(.*)$` being the default regex. With `buildozer $2 $3` being the command to be prompted to run?

Another thing to consider - as we continue to add more options to ibazel, is it worth adding a `tools/ibazel.rc` file? Similar to bazel's `tools/bazel.rc` file? As we add more options, typing them out everytime is cumbersome.

ittaiz:
One note re buildozer- the instructions from java rules are sometimes
incorrect so there was a discussion in bazel-discuss about replacing it
with a different tool.
We have someone internally which is starting to work on it next week.
This isn’t crucial since it’s not ready but just food for thought

achew22:
For the record, I think we should think long and hard before we make a regexp that grabs untrusted output from someone else's software and executes it on your machine. I think we will probably need to construct a very specific one for buildozer and any other tool that we want to run. Obviously this is susceptible to a malicious rules author, but I could pretty easily imagine someone outputting as error output a rule (non-maliciously):

`That feature isn't implemented yet but will fix it for real if you send me some $ rm -rf $BAZEL_CACHE/; as a temporary fix in the meantime`

Assuming line wrap at 80 characters, that string breaks the $ onto a newline, `$BAZEL_CACHE` isn't defined and becomes an empty string, and it is parsed as `rm -rf /` (assuming environment variable expansion).

jjudd:
I agree. However, the default is that output runner is disabled. When enabled, the default is to prompt, showing the user the command that will execute - with a default of No.

borkaehw:
Here is my proposal what to change.

I will change `-run_output` from type string to boolean. If the flag is provided, the default file location is `tools/bazel_fix_commands.json`. If the file is not found, default regex `^(buildozer) '(.*)'\\s+(.*)$` is applied.

In this way, users don't have to type file path in command line.

Is anyone against this change?

alexeagle:
I think the Bazel team is removing the tools directory as being special,
they are sending out changes to move bazelrc files.

borkaehw:
That is good to know! So what might be a better default path?

jjudd:
Yeah, Bazel moved `tools/baze.rc` to `.bazelrc`, but this broke quite a few people, so they are figuring out where to put things. There is also an issue to move `tools/bazel` somewhere else https://github.com/bazelbuild/bazel/issues/5788

Bazel seems to be moving things to the root of workspace. Any reason not to just make this a dotfile at the root of the workspace?

borkaehw:
OK, I made some changes
1. `-run_output` is a boolean flag, the default JSON path is at workspace root `.bazel_fix_commands.json`.
2. If file is not found, `buildozer` is the default command to look for.

achew22:
@borkaehw  I think .bazel_fix_commands.json is great. Thanks for doing that. Lemme trigger the buildkite run

borkaehw:
@achew22 It seems like we have passed all the checks. Is it ready to be merged?

ibazel/output_runner/output_runner.go
```diff
@@ -0,0 +1,189 @@
...
+func (i *OutputRunner) BeforeCommand(targets []string, command string) {}
+
+func (i *OutputRunner) AfterCommand(targets []string, command string, success bool, output *bytes.Buffer) {
+	if *runOutput == false || output == nil {
+		return
+	}
+
+	jsonCommandPath := ".bazel_fix_commands.json"
+	defaultRegex := Optcmd{
+		Regex: "^(buildozer) '(.*)'\\s+(.*)$",
```
achew22: Just as a note this is a regexp, as opposed to a pushdown automata, and so complex stuff like escaping single quotes is not going to work.There is some subtlty to correctly parsing bash that I think is elided in here. For example `buildozer "do stuff" //my:target` and `buildozer print //target` are not going to workeven though they are valid. I don't think it's worth the effort to construct a parser for this, but I think it would be worth indicating to rules authors in a comment.

achew22:
A couple of minor nits and then I'm very happy to merge and even do a release. I have time to do that this evening if you can get it turned before 6pm pacific

ibazel/command/default_command.go
```diff
@@ -54,29 +55,31 @@ func (c *defaultCommand) Terminate() {
...
-func (c *defaultCommand) Start() error {
+func (c *defaultCommand) Start() (*bytes.Buffer, error) {
 	b := bazelNew()
 	b.SetArguments(c.bazelArgs)
 
 	b.WriteToStderr(true)
 	b.WriteToStdout(true)
 
-	c.cmd = start(b, c.target, c.args)
+	outputBuffer, foo := start(b, c.target, c.args)
```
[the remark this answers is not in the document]
borkaehw: Let me know if it looks better.

ibazel/output_runner/output_runner.go
```diff
@@ -0,0 +1,189 @@
...
+		Command: "$1",
```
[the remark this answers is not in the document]
borkaehw: Good idea.

ibazel/output_runner/output_runner.go
JaredNeil: We are not parsing bash syntax at all. We are just parsing the build command output and constructing our own shell command from the captured groups. The output could just as easily be a compile or lint error with file names and line numbers.

borkaehw: I think we are not parsing arbitrary bash here. Rules authors may design specific command format, and since JSON is configurable, users can parse the commands they are looking for. Unless we change the buildozer commands, this default regex will work. Am I right? Also, where is a good place to leave a comment?

borkaehw:
Please take another look for the changes. Also, we are not sure why we fail Travis CI build.
```
/home/travis/.cache/bazel/external/com_github_fsnotify_fsnotify/BUILD.bazel:3:1: no such package
```
We don't think we change something related.

achew22:
I just tagged [v0.5.0](https://github.com/bazelbuild/bazel-watcher/releases/tag/v0.5.0) and released to [npm](https://www.npmjs.com/package/@bazel/ibazel).

Thanks so much for your contribution! I really look forward to the project growing with your support.

Globegitter:
@borkaehw @achew22 this feature looks awesome, is there any way to debug this? I am running a go binary in the following way: `ibazel -run_output_interactive -run_output run //something:go.binary` and have a `.bazel_fix_commands.json` in my workspace root but nothing seems to be happening when I add a new dep to my go code and ibazel/bazel throws an error about it. Even something silly like 
```
[
    {
        "regex": ".*",
        "command": "bazel"
    }
]
```

does not seem to trigger anything.

borkaehw:
@Globegitter Hi, thanks for the feedback. This feature is designed to pass in JSON format like this
```
[
	{
		"regex": "^(buildozer) '(.*)'\\s+(.*)$",
		"command": "$1",
		"args": ["$2", "$3"]
	}
]
```
you may take a look at `ibazel/output_runner/output_runner_test.json` for reference. I am not sure if it is because you are missing `args`. Would you try again, otherwise I would like to know what error did you get? Thanks.

Issue #117, opened 2018-01-26
ibazel watch stops working after 2 refreshes


alexeagle:
Originally reported at:
https://github.com/angular/angular/issues/21770

Has been observed by several engineers on the Angular team.

achew22:
Is it consistent on every other save it stops working? Is it consistent across editors? The other report says OS X (which I don't have one of unfortunately), does it happen on your linux workstations the same?

IgorMinar:
I'm able to repro it consistently across any shell/editor on MacOS. All of the repros of this that I'm aware were on MacOS.

achew22:
I have a work lap'n'top that is a mac. I'll test it out on there when I get back to the states. Sorry for slow response on this one -- it'll take me a couple of days to get a test rig set up.

alexeagle:
No problem, we're in this for the long term. +1 that I haven't observed it on Linux. Sorry this project is so platform-finicky

clintharrison:
Has anyone had a chance to look into this further? I'm running into the same issue, but if anyone has made any progress I'd appreciate not duplicating that work.

jjudd:
We're running into this on Ubuntu 16.04. We're going to dive in and attempt to fix it in the next few weeks. Anyone have any additional info on this? We think it might be related to the open files/inotify watches errors issues in this project, but that's a guess right now.

clintharrison:
I'm also on macOS, 10.12.6 currently. I unfortunately haven't had time to look into this, so I don't have any more details to provide :\

joprice:
This may affect adoption especially for web projects, where a livereload workflow is expected. Is there a working commit to use before this issue was introduced?

jjudd:
@joprice which OS are you using? We spent quite a bit of time trying to reproduce this on Ubuntu. WE found a bug in code we wrote that isn't on master, but otherwise couldn't reproduce this issue.

joprice:
Not sure if it's related, but I added a listener for errors on `SourceFileWatcher` and I see the following error after the first update:

`Error writing success to stdin: write |1: broken pipe`

```go
go func() {
    for err := range handler.SourceFileWatcher.Errors {
      fmt.Fprintf(os.Stderr, "Error %s\n", err)
    }
  }()
```

joprice:
[...]

PR #139, opened 2018-09-28
Add test reproducing vim backupcopy=no and kqueue failure

clintharrison:
This test reproduces the write behavior of vim with `backupcopy` set to both `yes` and `no`. `yes` is the default on Linux, and #10 fixed this case. Unforunately, on macOS, `yes` is not the default, and using `no` will cause ibazel to stop watching the file after a single file save when a kqueue-based file watcher is used. This is described in further detail in #117.

The fix for this might actually be the approach first used in #10 except for the `Rename` event, not just `Remove` (that is, waiting for vim to finish writing to the new file and re-adding the watcher on the original filename).

Otherwise, we may need to switch to an FSEvents-based implementation on macOS, which would require some upstream work on https://github.com/fsnotify/fsevents or waiting for https://github.com/fsnotify/fsnotify/issues/11 to be finished.

### DIFF
--- a/e2e/simple/simple_test.go
+++ b/e2e/simple/simple_test.go
@@ -1,6 +1,8 @@
 package simple
 
 import (
+	"fmt"
+	"io"
 	"os"
 	"runtime/debug"
 	"testing"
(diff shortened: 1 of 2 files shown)


achew22:
> The fix for this might actually be the approach first used in #10 except for the Rename event, not just Remove (that is, waiting for vim to finish writing to the new file and re-adding the watcher on the original filename).

Yep, I agree that adding a watcher on Rename is a reasonable way to deal with this. Implementing a new mode for fsnotify is beyond what I'm willing to do.

Thanks so much for sending this PR in. This should hopefully help with one of our longest standing issues. Big big props, I really can't thank you enough! 🎉

achew22:
Marking as manual CLA override because I pushed an extra commit on top of @clintharrison's. When I merge these I will squash and they will become one commit with @clintharrison as the author.

Issue #105, opened 2017-12-20
Add Windows support


benvanik:
Currently things seem pretty unhappy on Windows, and it'd be really awesome if support was added. fsnotify already supports Windows, which from a brief scan of the code looks like the most platform-specific thing.

* Windows 10
* Go 1.9.2
* Bazel 0.8.1

Attempting `bazel build //ibazel` from cmd leads to:
```
λ bazel build //ibazel
WARNING: C:/users/ben/appdata/local/temp/_bazel_ben/fa1bxbsu/external/com_github_bazelbuild_bazel_integration_testing/WORKSPACE:1: Workspace name in C:/users/ben/appdata/local/temp/_bazel_ben/fa1bxbsu/external/com_github_bazelbuild_bazel_integration_testing/WORKSPACE (@build_bazel_integration_testing) does not match the name given in the repository's definition (@com_github_bazelbuild_bazel_integration_testing); this will cause a build error in future versions
ERROR: D:/tools/bazel-watcher/ibazel/BUILD:30:1: no such package '@com_github_fsnotify_fsnotify//': failed to fetch com_github_fsnotify_fsnotify: 2017/12/19 18:02:56 --remote should be used with the --vcs flag. If this is an import path, use --importpath instead.
 and referenced by '//ibazel:go_default_library'
ERROR: Analysis of target '//ibazel:ibazel' failed; build aborted: no such package '@com_github_fsnotify_fsnotify//': failed to fetch com_github_fsnotify_fsnotify: 2017/12/19 18:02:56 --remote should be used with the --vcs flag. If this is an import path, use --importpath instead.
INFO: Elapsed time: 1.763s
FAILED: Build did NOT complete successfully (19 packages loaded)
```

This may be a `go_repository` issue.

If running from the msys bash the above doesn't trigger but instead nothing happens:
```
$ bazel build //ibazel
WARNING: C:/users/ben/appdata/local/temp/_bazel_ben/fa1bxbsu/external/com_github_bazelbuild_bazel_integration_testing/WORKSPACE:1: Workspace name in C:/users/ben/appdata/local/temp/_bazel_ben/fa1bxbsu/external/com_github_bazelbuild_bazel_integration_testing/WORKSPACE (@build_bazel_integration_testing) does not match the name given in the repository's definition (@com_github_bazelbuild_bazel_integration_testing); this will cause a build error in future versions
INFO: Analysed 0 targets (0 packages loaded).
INFO: Found 0 targets...
[...]

PR #144, opened 2018-10-07
Add Windows support

jchv:
This PR makes ibazel build and run on Windows. Tests do not pass on Windows, and Process Group is not tested at all (at least not directly) so it is definitely not quite ready. I am posting it mostly to get a litmus test on the approach as well as to potentially sync efforts.

Issue: #105.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,28 +1,41 @@
 ---
 platforms:
   ubuntu1404:
+    build_flags:
+    - "--build_tag_filters=-nolinux"
     build_targets:
     - "..."
     test_flags:
     - "--features=race"
+    - "--test_tag_filters=-nolinux"
     test_targets:
     - "..."
   ubuntu1604:
+    build_flags:
+    - "--build_tag_filters=-nolinux"
     build_targets:
     - "..."
     test_flags:
     - "--features=race"
+    - "--test_tag_filters=-nolinux"
     test_targets:
     - "..."
   macos:
+    build_flags:
+    - "--build_tag_filters=-nomacos"
     build_targets:
     - "..."
     test_flags:
     - "--features=race"
+    - "--test_tag_filters=-nomacos"
     test_targets:
     - "..."
   windows:
+    build_flags:
+    - "--build_tag_filters=-nowindows"
     build_targets:
     - "..."
+    test_flags:
+    - "--test_tag_filters=-nowindows"
     test_targets:
     - "..."
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -21,3 +21,8 @@ platforms:
     - "--features=race"
     test_targets:
     - "..."
+  windows:
+    build_targets:
+    - "..."
+    test_targets:
+    - "..."
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -37,5 +37,6 @@ platforms:
     - "..."
     test_flags:
     - "--test_tag_filters=-nowindows"
+    - "--experimental_enable_runfiles"
     test_targets:
     - "..."
--- a/e2e/common.go
+++ b/e2e/common.go
@@ -22,12 +22,13 @@ func GetPath(p string) string {
 	return path
 }
 
-var ibazelPath string
+var ibazelPath = getiBazelPath()
 
-func init() {
-	var err error
-	ibazelPath = GetPath(fmt.Sprintf("ibazel/%s_%s_pure_stripped/ibazel", runtime.GOOS, runtime.GOARCH))
-	if err != nil {
-		panic(err)
+func getiBazelPath() string {
+	suffix := ""
+	// Windows expects executables to end in .exe
+	if runtime.GOOS == "windows" {
+		suffix = ".exe"
 	}
+	return GetPath(fmt.Sprintf("ibazel/%s_%s_pure_stripped/ibazel%s", runtime.GOOS, runtime.GOARCH, suffix))
 }
--- a/e2e/ibazel.go
+++ b/e2e/ibazel.go
@@ -158,10 +158,7 @@ func (i *IBazelTester) build(target string, additionalArgs []string) {
 }
 
 func (i *IBazelTester) run(target string, additionalArgs []string) {
-	args := []string{
-		"--bazel_path=" + i.bazelPath(),
-		"--log_to_file=/tmp/ibazel_output.log",
-	}
+	args := []string{"--bazel_path=" + i.bazelPath()}
 	args = append(args, additionalArgs...)
 	args = append(args, "run")
 	args = append(args, target)
--- a/e2e/live_reload/BUILD
+++ b/e2e/live_reload/BUILD
@@ -10,6 +10,7 @@ bazel_go_integration_test(
         "//ibazel",
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/e2e/live_reload",
+    tags = ["nowindows"],
     versions = GET_LATEST_BAZEL_VERSIONS(),
     deps = [
         "//e2e:go_default_library",
--- a/e2e/output_runner/BUILD
+++ b/e2e/output_runner/BUILD
@@ -8,6 +8,7 @@ bazel_go_integration_test(
         "//ibazel",
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/e2e/output_runner",
+    tags = ["nowindows"],
     versions = GET_LATEST_BAZEL_VERSIONS(),
     deps = [
         "//e2e:go_default_library",
--- a/e2e/profiler/BUILD
+++ b/e2e/profiler/BUILD
@@ -10,6 +10,7 @@ bazel_go_integration_test(
         "//ibazel",
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/e2e/profiler",
+    tags = ["nowindows"],
     versions = GET_LATEST_BAZEL_VERSIONS(),
     deps = [
         "//e2e:go_default_library",
--- a/e2e/simple/BUILD
+++ b/e2e/simple/BUILD
@@ -10,6 +10,7 @@ bazel_go_integration_test(
         "//ibazel",
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/e2e/simple",
+    tags = ["nowindows"],
     versions = GET_LATEST_BAZEL_VERSIONS(),
     deps = [
         "//e2e:go_default_library",
--- a/ibazel/BUILD
+++ b/ibazel/BUILD
@@ -34,6 +34,8 @@ go_library(
         "ibazel.go",
         "lifecycle.go",
         "main.go",
+        "main_unix.go",
+        "main_windows.go",
         "source_event_handler.go",
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/ibazel",
--- a/ibazel/command/BUILD
+++ b/ibazel/command/BUILD
@@ -23,7 +23,10 @@ go_library(
     ],
     importpath = "github.com/bazelbuild/bazel-watcher/ibazel/command",
     visibility = ["//ibazel:__subpackages__"],
-    deps = ["//bazel:go_default_library"],
+    deps = [
+        "//bazel:go_default_library",
+        "//ibazel/process_group:go_default_library",
+    ],
 )
 
 go_test(
@@ -39,5 +42,6 @@ go_test(
     deps = [
         "//bazel:go_default_library",
         "//bazel/testing:go_default_library",
+        "//ibazel/process_group:go_default_library",
     ],
 )
--- a/ibazel/command/command.go
+++ b/ibazel/command/command.go
@@ -43,7 +43,7 @@ type Command interface {
 // common.
 func start(b bazel.Bazel, target string, args []string) (*bytes.Buffer, process_group.ProcessGroup) {
 	var filePattern strings.Builder
-	filePattern.WriteString("bazel_script_path")
+	filePattern.WriteString("bazel_script_path*")
 	if runtime.GOOS == "windows" {
 		filePattern.WriteString(".bat")
 	}
--- a/ibazel/command/command.go
+++ b/ibazel/command/command.go
@@ -22,12 +22,12 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
-	"syscall"
 
 	"github.com/bazelbuild/bazel-watcher/bazel"
+	"github.com/bazelbuild/bazel-watcher/ibazel/process_group"
 )
 
-var execCommand = exec.Command
+var execCommand = process_group.Command
 var bazelNew = bazel.New
 
 // Command is an object that wraps the logic of running a task in Bazel and
@@ -41,7 +41,7 @@ type Command interface {
 
 // start will be called by most implementations since this logic is extremely
 // common.
-func start(b bazel.Bazel, target string, args []string) (*bytes.Buffer, *exec.Cmd) {
+func start(b bazel.Bazel, target string, args []string) (*bytes.Buffer, process_group.ProcessGroup) {
 	var filePattern strings.Builder
 	filePattern.WriteString("bazel_script_path")
 	if runtime.GOOS == "windows" {
@@ -65,11 +65,8 @@ func start(b bazel.Bazel, target string, args []string) (*bytes.Buffer, *exec.Cm
 	// Now that we have built the target, construct a executable form of it for
 	// execution in a go routine.
 	cmd := execCommand(runScriptPath, args...)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-
-	// Set a process group id (PGID) on the subprocess. This is
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	cmd.RootProcess().Stdout = os.Stdout
+	cmd.RootProcess().Stderr = os.Stderr
 
 	return outputBuffer, cmd
 }
--- a/ibazel/command/command_test.go
+++ b/ibazel/command/command_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/bazelbuild/bazel-watcher/bazel"
+	"github.com/bazelbuild/bazel-watcher/ibazel/process_group"
 )
 
 var oldExecCommand = execCommand
@@ -36,7 +37,7 @@ func assertKilled(t *testing.T, cmd *exec.Cmd) {
 }
 
 func TestSubprocessRunning(t *testing.T) {
-	execCommand = func(name string, args ...string) *exec.Cmd {
+	execCommand = func(name string, args ...string) process_group.ProcessGroup {
 		return oldExecCommand("ls") // Every system has ls.
 	}
 	defer func() { execCommand = oldExecCommand }()
--- a/ibazel/command/default_command.go
+++ b/ibazel/command/default_command.go
@@ -18,15 +18,15 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"os/exec"
-	"syscall"
+
+	"github.com/bazelbuild/bazel-watcher/ibazel/process_group"
 )
 
 type defaultCommand struct {
 	target    string
 	bazelArgs []string
 	args      []string
-	cmd       *exec.Cmd
+	pg        process_group.ProcessGroup
 }
 
 // DefaultCommand is the normal mode of interacting with iBazel. If you start a
@@ -41,7 +41,7 @@ func DefaultCommand(bazelArgs []string, target string, args []string) Command {
 }
 
 func (c *defaultCommand) Terminate() {
-	if !subprocessRunning(c.cmd) {
+	if c.pg != nil && !subprocessRunning(c.pg.RootProcess()) {
 		return
 	}
 
@@ -50,9 +50,9 @@ func (c *defaultCommand) Terminate() {
 	// send to the PGID, send the signal to the negative of the process PID.
 	// Normally I would do this by calling c.cmd.Process.Signal, but that
 	// only goes to the PID not the PGID.
-	syscall.Kill(-c.cmd.Process.Pid, syscall.SIGKILL)
-	c.cmd.Wait()
-	c.cmd = nil
+	c.pg.Kill()
+	c.pg.RootProcess().Wait()
+	c.pg = nil
 }
 
 func (c *defaultCommand) Start() (*bytes.Buffer, error) {
@@ -63,12 +63,12 @@ func (c *defaultCommand) Start() (*bytes.Buffer, error) {
 	b.WriteToStdout(true)
 
 	var outputBuffer *bytes.Buffer
-	outputBuffer, c.cmd = start(b, c.target, c.args)
+	outputBuffer, c.pg = start(b, c.target, c.args)
 
-	c.cmd.Env = os.Environ()
+	c.pg.RootProcess().Env = os.Environ()
 
 	var err error
-	if err = c.cmd.Start(); err != nil {
+	if err = c.pg.Start(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error starting process: %v\n", err)
 		return outputBuffer, err
 	}
@@ -83,5 +83,5 @@ func (c *defaultCommand) NotifyOfChanges() *bytes.Buffer {
 }
 
 func (c *defaultCommand) IsSubprocessRunning() bool {
-	return subprocessRunning(c.cmd)
+	return c.pg != nil && subprocessRunning(c.pg.RootProcess())
 }
--- a/ibazel/command/default_command.go
+++ b/ibazel/command/default_command.go
@@ -52,6 +52,7 @@ func (c *defaultCommand) Terminate() {
 	// only goes to the PID not the PGID.
 	c.pg.Kill()
 	c.pg.Wait()
+	c.pg.Close()
 	c.pg = nil
 }
 
--- a/ibazel/command/default_command.go
+++ b/ibazel/command/default_command.go
@@ -51,7 +51,7 @@ func (c *defaultCommand) Terminate() {
 	// Normally I would do this by calling c.cmd.Process.Signal, but that
 	// only goes to the PID not the PGID.
 	c.pg.Kill()
-	c.pg.RootProcess().Wait()
+	c.pg.Wait()
 	c.pg = nil
 }
 
--- a/ibazel/command/default_command_test.go
+++ b/ibazel/command/default_command_test.go
@@ -16,18 +16,16 @@ package command
 
 import (
 	"os"
-	"os/exec"
-	"syscall"
 	"testing"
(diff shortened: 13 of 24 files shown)


meteorcloudy:
The tests are failing with
```
2018/10/09 00:19:16 could not change to test directory: chdir C:\users\b\_bazel_b\dnil4uob\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\bazel\windows_amd64_stripped\go_default_test.exe.runfiles\__main__\bazel: The system cannot find the file specified.
```

On Windows we don't have runfiles tree enabled by default, it's recommend to use a runfiles library to access all the data files.
You can also try to use `--experimental_enable_runfiles" flag, maybe that will make the test pass.

achew22:
```
bazel test --show_progress_rate_limit=5 --curses=yes --color=yes --keep_going --jobs=32 --build_event_json_file=D:\temp\tmp3dc_dunz\test_bep.json --experimental_build_event_json_file_path_conversion=false --announce_rc --sandbox_tmpfs_path=/tmp --experimental_multi_threaded_digest --flaky_test_attempts=3 --build_tests_only --local_test_jobs=8 --test_tag_filters=-nowindows --experimental_enable_runfiles ...
  | INFO: Options provided by the client:
  | Inherited 'common' options: --isatty=0 --terminal_columns=80
  | INFO: Options provided by the client:
  | Inherited 'build' options: --python_path=C:/python3/python.exe
  | INFO: Reading rc options for 'test' from d:\b\bk-worker-windows-java8-w5ct\bazel\bazel-watcher\tools\bazel.rc:
  | Inherited 'build' options: --workspace_status_command=tools/workplace_status.sh
  | ERROR: building runfiles is not supported on Windows
  | ERROR: Build options are invalid
  | INFO: Elapsed time: 0.398s
  | INFO: 0 processes.
  | ERROR: Couldn't start the build. Unable to run tests
  | FAILED: Build did NOT complete successfully (0 packages loaded)
  | At least one test failed or was flaky.

```

achew22:
@meteorcloudy Do you know what's up with the `  | ERROR: building runfiles is not supported on Windows` error output from Bazel on Windows?

meteorcloudy:
Oh, we are still using Bazel 0.17.2 on CI, which doesn't support building symlink runfiles tree on Windows. I'll ping you when it's updated.

meteorcloudy:
Oh, we are still using Bazel 0.17.2 on CI, which doesn't support building symlink runfiles tree on Windows. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Looks like we're still using Bazel 0.17.2 on CI, but building runfiles tree on Windows only works with Bazel 0.18.0 or later. I'll ping you when it's upgraded.

meteorcloudy:
Hi @achew22, Bazel has been upgraded to 0.18.0 on CI, I'm rerunning the test on Windows

meteorcloudy:
Three tests are passing now
```
//bazel:go_default_test                                                  PASSED in 0.7s
//ibazel:go_default_test                                                 PASSED in 0.8s
//ibazel/output_runner:go_default_test                                   PASSED in 0.7s
```
The rest are failing with:
```
panic: unable to find file "ibazel/windows_amd64_pure_stripped/ibazel"
```
Is it supposed to be `ibazel.exe` on Windows?

meteorcloudy:
Another question, does ibazel itself needs the runfiles symlink tree to run? Or it's jsut needed for the tests.

achew22:
> Another question, does ibazel itself needs the runfiles symlink tree to run? Or it's jsut needed for the tests.

No, ibazel takes all its data dependencies and converts them to string variables in a [go_embed_data](https://github.com/bazelbuild/rules_go/blob/master/extras/embed_data.bzl) rule which means I shouldn't have any production runtime deps. I wrote this in go so it could be statically linked and distributed without any `.so`s.

meteorcloudy:
Hi @jchv , I debugged ibazel on Windows with your PR, and found `ibazel run` still doesn't work on Windows and it requires fixes from both Bazel and ibazel.

From Bazel side:
We need to fix `--script_path` on Windows, because it always writes a bash script, but we need a batch script for Windows.
It is already done at https://github.com/bazelbuild/bazel/commit/4634c20a858af4abb91e08f1a3e810acccc0e4a6. Please build a bazel from HEAD at use it for ibazel

From ibazel side:

- The temporary script generated by `--script_path` should have a `.bat` extension, otherwise you cannot execute it
See my change https://github.com/meteorcloudy/bazel-watcher/commit/ea225fe0a7a9a9ff7e1939399a7cd5e593419fc7#diff-d50bab4ab597fcef4ca292fd79c1359c

- After fixing the extension, I was getting an "The handle is invalid" error, which I believe something is wrong in [process_group_windows.go](https://github.com/bazelbuild/bazel-watcher/pull/144/files#diff-77979a78d8911c97eb64f2d781a155cb)

John, since you authored this change, can you help debug and make `ibazel run` work with Bazel@HEAD?

jchv:
Sure, I'll take another look. I had been using `ibazel test` which was at some point working for me, so I didn't catch that `ibazel run` wasn't working.

There's probably bugs in `process_group_windows.go`, but those aside I also need to go and clean up the error handling, which is probably not sufficient at the moment.

jchv:
Sorry for lack of updates. I went at it again this weekend, but I'm sad to say I don't know how to progress.

So, I updated to Bazel HEAD today, and ran into some new issues, specifically in @io_bazel_rules_go//:go_context_data.
```
Starlark APIs accessing compilation flags has been removed. Use the new API on cc_common (see --incompatible_disable_legacy_flags_cc_toolchain_api onhttps://docs.bazel.build/versions/master/skylark/backward-compatibility.html#disable-legacy-c-toolchain-api for migration notes).
```

I re-enabled it. Then got this:
```
Starlark APIs accessing compilation flags has been removed. Use the new API on cc_common (see --incompatible_disable_legacy_flags_cc_toolchain_api onhttps://docs.bazel.build/versions/master/skylark/backward-compatibility.html#disable-legacy-c-toolchain-api for migration notes).
```

Sadly, looks like this one can't be re-enabled.

I was hoping maybe there was a version of Bazel I could use before this deprecation, so I went ahead and built bazel at ea225fe0a7a9a9ff7e1939399a7cd5e593419fc7. This didn't seem to work either, same exact issues. I continued on with this version of Bazel.

The problem was fixed in io_bazel_rules_go 0.6.1. I tried that:
```
ERROR: C:/bazel/n7nxee47/external/com_google_protobuf/BUILD:70:1: C++ compilation of rule '@com_google_protobuf//:protobuf_lite' failed (Exit 2)
cl : Command line error D8021 : invalid numeric argument '/Wwrite-strings'
```

I traced it down to this:
```python
COPTS = select({
    ":msvc" : MSVC_COPTS,
    "//conditions:default": [
        "-DHAVE_PTHREAD",
        "-Wall",
        "-Wwrite-strings",
        "-Woverloaded-virtual",
        "-Wno-sign-compare",
        "-Wno-unused-function",
        # Prevents ISO C++ const string assignment warnings for pyext sources.
        "-Wno-writable-strings",
    ],
})
```

`:msvc` is 
```
config_setting(
    name = "msvc",
    values = { "compiler": "msvc-cl" },
)
```

OK, so maybe if I explicitly specify `--compiler=msvc-cl` it will work?

Well, I got past that issue. But then:

```
ERROR: C:/projects/go/src/github.com/bazelbuild/bazel-watcher/npm/BUILD:17:1: Executing genrule //npm:package failed (Exit 2)
panic: The version string was not overriden. Please rebuild with --stamp
```

Adding `--stamp` does not help.

This is where I am currently at. Sadly, trying to get anything working on Windows is an exercise in frustration at the moment. I suppose in order to progress further, there are more upstream changes that will need to be made, and right now I do not have the expertise to trace this any further.

I might try to take another look at the problem later, but if anyone has any insight as to how I can move forward it would be appreciated.

achew22:
@jchv, Thanks so much for the work you've done here! I can't tell you how much I appreciate you putting in so much effort.

I just spent the last hour doing a bit of cleanup around the repo and I think some of these problems should be fixed for you. For example, I just merged #153 which I believe fixes the --stamp problem. It turns out that Windows doesn't have a .sh interpreter so you have to use python. Who knew? I also moved the .bazelrc file into the correct location so that if you're using a modern bazel it will behave correctly (#154). I upgraded rules_go past the version you specified to 0.16.1 in (#152).

I think this fixes a few more of the problems but there are almost certainly more lingering around.

> I might try to take another look at the problem later, but if anyone has any insight as to how I can move forward it would be appreciated.

@meteorcloudy is the person on the Bazel team who is responsible for Bazel on Windows and he has been really helpful to me in debugging issues. I think I fixed the stamp problem but if you see something else please comment on here and we can figure it out together!

Really, thank you so much! I don't think I could have done this.

achew22:
https://github.com/golang/go/issues/5615 might have some more info that's useful to us here.

> That is expected - if you start process, you get permission to kill it by default.
> I think, if there are no objections, I will try to do what I suggested in #1. I won't be
> able to try for a few days.

achew22:
I'm wondering if this might be a problem that is finicky enough that we don't care to solve it. Poking around at the error I think it comes from the test case triggering an event in iBazel such that it needs to restart the process (and thus kills it). This wouldn't be a problem normally, but we are using process groups which means everything is hard. Maybe we can enhance our permissions when we start the process. I will now do something that I never thought I would ever do -- link to some MS win32api documentation. *Shudder*

https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-openprocess

Looks like there is a permission that can be passed at construction time where you can allow TERMINATE_PROCESS.

We may have to use the windows speific golang api to do this https://godoc.org/golang.org/x/sys/windows which exports that value.

achew22:
https://godoc.org/golang.org/x/sys/windows#OpenProcess is the API that takes that permission

achew22:
I was hoping for a windows specific SysProcAttr, but I don't think we are so lucky

https://godoc.org/golang.org/x/sys/windows#SysProcAttr

achew22:
Looks like there are two problems

1. The [`ibazel.go` e2e helper](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/ibazel.go#L103) isn't properly killing subtasks in Windows
1. The e2e helper is not hooking up reading `stdout` properly which means that [`GetOutput()`](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/live_reload/live_reload_test.go#L80) isn't working. One thing to check is to expand the [ExpectedOutput](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/ibazel.go#L55:24) method to panic if the expected output isn't received. If that happens it feels like printing out all the process information is not a terrible next step.

Control flow that is causing this problem:

 1.  [Enter test](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/live_reload/live_reload_test.go#L58)
 1.  [Launch iBazel](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/live_reload/live_reload_test.go#L58)
 1.  [Defer a kill of iBazel](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/live_reload/live_reload_test.go#L77)
 1.  [GetOutput() fails](https://github.com/bazelbuild/bazel-watcher/blob/f897ad32f6c5a4fb7934ac218c1257f6751fd51a/e2e/live_reload/live_reload_test.go#L80) I think this is the key right here.
 1.  The defer is executed which then gives us the following stacktrace

```
goroutine 19 [running]:
testing.tRunner.func1(0xc00013c100)
	GOROOT/src/testing/testing.go:792 +0x38e
panic(0x6b1820, 0xc0000046e0)
	GOROOT/src/runtime/panic.go:513 +0x1c7
github.com/bazelbuild/bazel-watcher/e2e.(*IBazelTester).Kill(0xc000473e10)
	e2e/ibazel.go:105 +0x6a
runtime.Goexit()
	GOROOT/src/runtime/panic.go:397 +0xfd
testing.(*common).FailNow(0xc00013c100)
	GOROOT/src/testing/testing.go:590 +0x40
testing.(*common).Fatal(0xc00013c100, 0xc000473d10, 0x1, 0x1)
	GOROOT/src/testing/testing.go:628 +0x76
github.com/bazelbuild/bazel-watcher/e2e/live_reload.TestLiveReload(0xc00013c100)
	e2e/live_reload/live_reload_test.go:84 +0x1008
testing.tRunner(0xc00013c100, 0x7195c8)
	GOROOT/src/testing/testing.go:827 +0xc6
created by testing.(*T).Run
	GOROOT/src/testing/testing.go:878 +0x35a
```

achew22:
With even more staring at the log output:

```
    ibazel.go:71: Expected iBazel output after 10s to be:
        Wanted [Live reload url: http://.+:\d+], got []
```

Looks like there is something hinky there. On the plus side, it is actually running in CI now. It's really late at night for me right now so I'm going to go to sleep and contemplate this in the morning. STDOUT isn't hooked up in Windows? Is that possible?

meteorcloudy:
@achew22 Thanks for making the `--workspace_command` work on Windows! We also need to fix the script file [extension](https://github.com/bazelbuild/bazel-watcher/blob/master/ibazel/command/command.go#L43) with 
```
	suffix := ""
	if runtime.GOOS == "windows" {
		suffix = ".bat"
	}
```

@jchv Yes, the protoc compile failure can be work around by `--compiler=msvc-cl`. We have to update protobuf version to include fixes from @scentini https://github.com/protocolbuffers/protobuf/commits?author=scentini

With addition fixes from @achew22 , I think you'll be able to move forward now. Please ping me if you have any other question.

achew22:
I just did the `.bat` suffix code in https://github.com/bazelbuild/bazel-watcher/pull/158

jchv:
So I am chipping away at various problems. I plan on looking at the process group specific issues again soon, but I want to make sure I at least visit most of the obvious problems I hit.

Another nasty one is with workspace status command. 

Under most UNIX-like OSes, this makes sense:
```
build --workspace_status_command=tools/workplace_status.py
```
...because `workspace_status.py` carries a shebang:

```
#! /usr/bin/env python3
```

Of course, Windows doesn't look at these, and that won't work. What actually happens is, workspace_status.py opens in my editor (Visual Studio Code.) Somehow this went unnoticed by me until today. Either way, now that I've noticed it, it explains the `--stamp` errors.

On the other hand, THIS does work on Windows:

```
build --workspace_status_command="python tools/workplace_status.py"
```

...and it *seems* to fix the `--stamp` errors.

I'm kind of curious to see if this solution also works under Linux, since I'm sure the Command implementation is vastly different for Windows. I guess we'll see. The approach of explicitly calling Python seems like a winning plan to me, regardless - at least until a better solution than `--workspace_status_command` exists.

---

Next up, I noticed that there was an attempt to fix processgroup Kill, but I think it didn't work, so I reverted it locally. This lead to:

```
panic: TerminateProcess: Access is denied.
```

I eventually realized that this was actually occurring in a completely different context, `IBazelTester.Kill`. Turns out it was also a red herring - the PID it was trying to kill didn't exist, because the process exited immediately trying to save its logs to `/tmp`: `--log_to_file=/tmp/ibazel_output.log`

It seems this was mostly for debugging, so I just removed it for now. The tests now felt like they were actually doing something, but a Windows Firewall dialog popped up for a few ibazel instances, which could be problematic for automated testing. They all failed for the same reason: ibazel was still terminating immediately. I found out why pretty quickly: the actual bazel binary was an bash script without an extension. Another thing you can do in a UNIX-like with execute bits, but not really on Windows.

So I think the next thing I need to work on for e2e tests is getting the Bazel Go integration testing library working on Windows. I'm not sure what the use in the shell script is versus just calling bazel-real directly, I'll have to investigate further.

---

Other than that, I still need two flags to get this far in testing: `--experimental_enable_runfiles --compiler=msvc-cl` - The former I've already been working on, but the latter I would also like to attack. Probably possible to replace the custom `config_setting` with one of the built-in `msvc` conditions, though thus far I've had issues with trying to do that.

jchv:
Please bear with me while the process group code devolves into a bit of chaos. It turns out the Win32 APIs are a bit more complicated than I thought.

It turns out:
- When you create a suspended process, you need the main thread handle to unsuspend it. Go discards this handle, so it's not accessible. At this point you can either duplicate all of the logic in cmd, or you can try to recover this information.
- I went with the approach to recover this information. It seems like `exec.Cmd` probably goes through a lot of trouble to deal with things like I/O and argument escaping, and I don't think duplicating it is a good idea.
- _The API call you need to get the threads for a process on Windows is `GetProcessThreads`_. Just kidding, that's not a real Win32 API, the ones we need are named `CreateToolhelp32Snapshot`, `Thread32First` and `Thread32Next`.
- `CreateToolhelp32Snapshot` has a Process ID parameter. It does not work for iterating threads threads.
- Therefore... you have to list every single thread that you have access to see in order to find the one you want.
- Once you're done and want to wait for the process group to finish, _you can use the standard `WaitForSingleObject` to wait until the process group is finished executing_. Nope, kidding. You have to:
    - Create an IO port with `CreateIoCompletionPort`
    - Use `SetInformationJobObject` with `JobObjectAssociateCompletionPortInformation` to associate the IO port with the job
    - Use `GetQueuedCompletionStatus` to get the completion status
    - Keep doing that until it returns the one that's actually for your job and event. Somehow, it returns things other than that too, so you will end up exiting early if you don't loop this.

That's where I'm at right now. I haven't actually tried the `ibazel` command in a while so for all I know it could be pretty broken, but at the very least some of the tests are now passing. Right now it's getting hung on a weird issue in `TestNotifyCommand`:

```
bazel test --test_arg=-test.v --experimental_enable_runfiles --compiler=msvc-cl //ibazel/command:go_default_test --test_output=all
```
```
--- FAIL: TestNotifyCommand (0.06s)
    notify_command_test.go:77: There are no more files.
    notify_command_test.go:83: Not equal.
        Got:  
        Want: IBAZEL_BUILD_COMPLETED SUCCESS
        IBAZEL_BUILD_COMPLETED FAILURE
        IBAZEL_BUILD_COMPLETED SUCCESS
FAIL
```

Line 77 is a call to ProcessGroup.CombinedOutput(). I have no idea what the "There are no more files" error means, but I would bet it's coming from Windows.

Personal TODO for this PR:
- [ ] Re-evaluate the design of `ProcessGroup`. It's starting to look like exposing `exec.Cmd` is a bad idea, and instead the `ProcessGroup` implementation should do all of the work. This is unfortunate in my opinion.
- [ ] Improve syscall usage. It would make the code a lot cleaner and easier to understand to separate the Win32 syscall wrappers out of the call sites and into helper functions
- [ ] Improve error handling. For the most part it turns out errno is sufficient by default, however I should handle the return value on a case-by-case basis.
- [ ] Fix handle leaks. I'm basically not closing handles right now. There's a lot of handle leaks.
- [ ] Look into dropping `exec.Cmd` for the Windows version of `ProcessGroup`. It would probably be worse to do that than it is to do what I am now, but maybe it's worth looking into just to be sure.

achew22:
> I have no idea what the "There are no more files" error means, but I would bet it's coming from Windows.

Is it possible that's a standin for a file not found error effectively? You removed the line where it writes stdout to file.

https://github.com/bazelbuild/bazel-watcher/pull/144/commits/8d0ca36d1aee1b8131cfd902862198fe02924744#diff-f44f3824181c55aee28e011bdcc40938R110

jchv:
Hmm, possible. My current supposition is that it has something to do with the plumbing in cmd.Exec for file descriptors being passed to stdout/stderr, because I think that error is actually coming from CombinedOutput. I'll try to do some more digging when I get back home.

jchv:
- Rebased everything to latest rev and force pushed to clean up commit log a bit.
- Finally have made progress, with at least one new test passing. Next thing I may have to do is poke at the Bazel integration test library for Go so that it can work on Windows.

achew22:
@meteorcloudy is it possible to add @jchv to the list of ppl who don't need to get their PR's verified? It would be nice for him to be able to get feedback from CI without having to go through me.

meteorcloudy:
@jchv Thanks for working on this again! 

The Bazel CI now doesn't requires a Bazel member to trigger the presubmit. If you rebase your branch, it should be able to run presubmit automatically.

Please tell me if you encounter any issue, I'd be glad to help!

meteorcloudy:
@jchv Perfect, thank you!

jchv:
Alright, now it seems to be working for both `run` and `test`, although I haven't tested extensively.

Work done since last time:
- Rebased everything against master. Minimal conflicts, all went well.
- Replaced horrible system thread iteration in `process_group` with `NtResumeProcess` function. Technically, this function is not part of the public Win32 API, but so much software uses it that it may as well be. The best path forward to get rid of this call is to patch Go itself. It looks like Job objects have been on the radar for a few years, so maybe eventually we'll just get Job objects directly? https://github.com/golang/go/issues/17245#issuecomment-352611923
- Patch to start routine to fix batch filename being wrong. Windows actually needs the extension to end in `.bat` otherwise it will not pick up the correct file associations to be executable. Someone already made sure `.bat` was ended, but the string passed to `TempFile` did not contain an asterisk, so the random numbers came after the file extension.

The checklist above still applies mostly, but also:

 - [ ] Still need to get e2e tests working. That work depends on https://github.com/bazelbuild/bazel-integration-testing/pull/96
 - [ ] Figure out why `//ibazel/output_runner:go_default_test` is failing.
 - [ ] I hardcoded some absolute paths in the command test because I found you can truly not depend on anything. When executing in msys2, Bazel tests can't seem to find Windows binaries. What would be even better than trying to resolve this for just Windows, would be if the commands themselves were Bazel targets brought in with a data dependency.

achew22:
@meteorcloudy, I'm not sure that's the case. It looks like the CI server isn't handling things correctly. It only ran 4 of the tests and none of the e2e tests were included. I'm aware that it is due to an incompatible change in Bazel, but it really seems like that should result in a CI failure.

```
INFO: 4 processes: 4 local.
//bazel:go_default_test                                                  PASSED in 0.6s
//ibazel:go_default_test                                                 PASSED in 0.7s
//ibazel/command:go_default_test                                         PASSED in 0.6s
//ibazel/output_runner:go_default_test                                   PASSED in 0.6s
 
Executed 4 out of 4 tests: 4 tests ``

meteorcloudy:
@achew22 Yes, that is definitely a Bazel bug, filed https://github.com/bazelbuild/bazel/issues/7115

To make the test work again, the fastest way is to fix bazel_integration_testing and upgrade it in bazel-watcher.

As for ibazel on Windows, I manually tested `ibazel run` and `ibazel test` on Windows, they both work quite well!

meteorcloudy:
@achew22 Can you help fixing bazel_integration_testing? I'm currently busy with other CI breakages.

achew22:
Did the e2e tests pass on your box?

meteorcloudy:
> Still need to get e2e tests working. That work depends on bazelbuild/bazel-integration-testing#96

As @jchv pointed out, we need to make bazel-integration-testing work on Windows first

meteorcloudy:
The e2e test are not even running on Linux, I think that's due to the incompatible change. Let's not merging this PR until we enable them again.

meteorcloudy:
@jchv Can you rebase to HEAD so that we can trigger the presubmit again.

achew22:
@meteorcloudy I was able to trigger this from my side. Thanks for pinging the PR

meteorcloudy:
@achew22 Thanks! Looks like the Windows failures are expected, e2e tests are failing because bazel_integration_test doesn't work on Windows yet.

jchv:
I'm interested in getting the e2e tests working, but in the meantime should we mark the tests to not run on Windows? There's really no need to rush to get this PR ready to merge imo, but some Windows support is probably better than none, and it seems movements to the other projects may take a long time.

(I still have a couple of other things to fix before I think it could possibly be reviewed and merged, namely the tests are still hacky.)

achew22:
@jchv TBH, I'm pretty reluctant to release into windows without any real test coverage. I know that a small amount comes in from the tests in `//ibazel/...` but if someone were to file a bug against me on Windows I would be shooting in the dark as to what happened.

Since I can only release to windows one time and I think there will be a flood of people (50-100 ppl), I don't want to leave a bad taste in their mouth for their first, and probably only if things go poorly, experience with ibazel.

jchv:
SGTM. Maybe to help reduce code rot I could at least split some of the less Windows-related work into other PRs to reduce the surface area, though. I am admittedly paranoid of having a large and invasive rebase down the line if it takes as long as I'm imagining to fix the upstream e2e test problems, since it's pending on something that seems to have been stalled for a while.

meteorcloudy:
@achew22 Merging this PR doesn't mean Windows support is ready and we have to do a release for Windows, right? We can still keep https://github.com/bazelbuild/bazel-watcher/issues/105 open until e2e tests are also enabled for Windows.

I think merging this change now could reduce the rebasing work @jchv had to do. Also, we'll at least have some test coverage on Bazel CI to prevent any breakage from Bazel itself. I just enabled Bazel watcher in downstream and it already caught a [Bazel remote execution bug](https://github.com/bazelbuild/bazel/issues/7160).

achew22:
You two are both totally correct and I'm 100% wrong. I'm going to tag the tests as `nowindows` and merge it then we can add the release of the exe in another pr when the e2e is good to go.

achew22:
Dear CLA bot, @jchv and @achew22 are both Google employees and we have both signed the CLA.

@jchv, thanks so much for all your hard work on this! I'm really excited to get to a place where we are releasing windows binaries.

PR #145, opened 2018-10-08
Support Bazel --output_groups flag

schroederc:


achew22:
@schroederc, thanks so much for your contribution!

PR #147, opened 2018-10-19
Support Bazel --keep_going flag

schroederc:


achew22:
@schroederc, thanks so much for your continued contribution!

PR #153, opened 2018-11-03
Switch to python for workplace_status script

achew22:


.bazelrc
```diff
@@ -1,4 +1,4 @@
-build --workspace_status_command=tools/workplace_status.sh
+build --workspace_status_command=tools/workplace_status.py
```
meteorcloudy: It's better to use `build --workspace_status_command="python tools/workplace_status.py"` because `.py` is not an executable extension on Windows by default.

achew22: Could I have you add a section to the Bazel docs on windows compatibility for this? This seems like a thing that is worth documenting. Do you know if those are open sourced yet?

meteorcloudy: Hmm, I'm also having the same concern as you do, that doing this will add python as a dependency for the project, but I couldn't think of a better solution yet. I prefer to discuss with my teammate (who's on vacation) first.

PR #163, opened 2018-11-12
Call bazel run directly.

Globegitter:
We started using [direnv](https://direnv.net/) for our projects and I added an auto-loading feature by providing a `tools/bazel` binary which then uses `direnv` to load the environment variables of the correct project and then just exports it before calling `$BAZEL_REAL`. But because bazel-watcher does the `script_path` thing any variables exported in the `tool/bazel` script are not available to the application that is being run.

I assume that this workaround was done of the way bazel used to invoke binaries. But since `--direct_run` was made the default a few versions ago there should be no issues anymore just calling `bazel run ...` directly. Or is there anything else I am missing?


Globegitter:
Will of course look into fixing the test but is that behaviour & implementation reasonable @achew22 ?

Globegitter:
Yep exactly that is the thought. Yeah the e2e tests have failed a bunch of times locally but good that ci is happy.

achew22:
Interesting that they failed locally. Do you know in what way they were failing?

Globegitter:
I can check again tomorrow but it was quite non-deterministic. Rerunning them a few times and then at some point all where passing.

ibazel/command/command.go
```diff
@@ -41,37 +37,18 @@ type Command interface {
...
+	args = append([]string{"run", target}, args...)
+	cmd := execCommand(*bazel.BazelPath, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
 	// Set a process group id (PGID) on the subprocess. This is
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
-	return outputBuffer, cmd
+	return nil, cmd
```
achew22: I think that by not returning the buffer [`output_runner`](https://github.com/bazelbuild/bazel-watcher/tree/master/ibazel/output_runner) is not going to be able to work. Can you help me understand why that isn't the case if you believe so?

ibazel/command/command.go
```diff
@@ -41,37 +37,18 @@ type Command interface {
...
-
-	// Start by building the binary
-	_, outputBuffer, _ := b.Run("--script_path="+tmpfile.Name(), target)
-
-	runScriptPath := tmpfile.Name()
-
-	// Now that we have built the target, construct a executable form of it for
-	// execution in a go routine.
-	cmd := execCommand(runScriptPath, args...)
+func start(target string, args []string) (*bytes.Buffer, *exec.Cmd) {
```
achew22: I think you should continue to use the bazel abstraction here

Globegitter: I was considering that but then it seemed the way the process is called is a bit different as ctrl+c did not work properly to interrupt the process. So I thought that maybe there is a reason why the binary was called with its own subprocess call. But can look into it in more detail to understand what's going on.

ibazel/command/command.go
Globegitter: Ah yeah that is a good point. I have only tried using it once. Did not immediately work and have not tried since. Also since autogazelle was released it is not that important for now for our bazel uses. Either way it should not break. Will also take a look.

achew22:
@Globegitter how can I help get this merged?

Globegitter:
@achew22 I have not had much time to look into this the last few days, but hopefully will find some time early next week. I did give it one shot to switch `execCommand` over to

```
cmd, outputBuffer, _ := b.Run(args...)
```

but somehow that does not work - if you have any idea of the top of your head what else would need to be changed that would be useful. Otherwise I'll just have to spend more time figuring out what exactly is happening.

Globegitter:
@achew22 I have pushed the progress I have made so far, maybe you can take a look and give me some pointers of where I am going wrong. Will still investigate and try and figure out what is going on whenever I have time.

achew22:
I just submitted #198 which adds some rudimentary test for output running. Feel free to use those as acceptance tests. If you get them to pass then I'm happy 😁

Globegitter:
I had more and more of a look at this and I am not sure if it is worth it anymore to go down this route. The main problem is really to get things like the outputrunner to work because they depend on the output of `bazel build` (which is done synchronously) and then the binary is started in a go routine. Where-as if we use `bazel run` we would have to start it in a go routine immediately and so we can not wait for any output and we would have to add some logic to watch the output and in some way or another stream that to things like the output runner. Right now the way we worked around this is by utilising the `bazel` script in our workspace. We added a custom `--watch` flag and a `watch` command which start `ibazel` directly and so we already have the right environment available.

The other change I was thinking off and I might submit a PR for that still is that we could get the environment from the `bazel build` command pass it over to the run of the binary that would also fix our particular issue without having to make use of the `bazel` script in the workspace.

The other idea that I had that could be interesting if ibazel would also call the `bazel` script or if it would have its own `ibazel` script it would call in the workspace if it exists. But also not sure if there really are many use case for this to be worth it.

achew22:
Just for my education purposes I'm kind of curious about your set up. Feel free to ignore if you want.

> I had more and more of a look at this and I am not sure if it is worth it anymore to go down this route. The main problem is really to get things like the outputrunner to work because they depend on the output of `bazel build` (which is called synchronously) and then the binary is started in a go routine. Where-as if we use `bazel run` we would have to start it in a go routine immediately and so we can not wait for any output and we would have to add some logic to watch the output and in some way stream that to things like the output runner.

As a strawman solution to this, WDYT about `bazel build`ing the target synchronously then `bazel run`ing the target async? That would provide similar effects and have only a small amount of additional time overhead.

> Right now the way we worked around this is by utilising the `bazel` script in our workspace. We added a custom `--watch` flag and a `watch` command which starts `ibazel` directly and so we already have the right environment set-up when calling ibazel and all is good.

So you created a script `bazel` at the path `tools/bazel` which automatically invokes `ibazel` but with some additional environment variables that make everything work. Another thing we could consider doing is creating a `before_run` and `after_run` flag. I presume it's possible to determine the environment variables of a subprocess and we could pass those down into the runner script? That is a hook I've wished for a few times but not quite enough to implement it yet. Maybe this is a good time to do it?

Issue #135, opened 2018-08-16
Watching for files being added


Globegitter:
When using globs, e.g. `data/**/*.txt` it would be nice if ibazel could watch this directory for new file additions and then rebuild. It would be even nicer if one could also specify to just watch for file additions, as e.g. for runfiles there is often no need to actually rebuild and the binary can just pick up the changed files due to it being a symlink.

Not sure how difficult this would be to implement, but with some thoughts/pointers on this I would be happy to look into this as well when I get some time for it.

achew22:
If you edit a BUILD file that is part of the sources, it should trigger a requery and find files. I looked into the idea of detecting these globs and doing my own globbing. https://github.com/google/skylark/issues/109 is where I asked about it publicly and I believe the analysis in there is correct. It would be great, but unfortunately it is going to take a massive quantity of code to do it. If you have theories on ways we can achieve this, I would *LOVE* to hear your input.

Thanks!

Globegitter:
Yeah, doing it in the way discussed in the ticket would be great i.e. that bazel watcher could parse the build files in parallel or get that information from bazel in a streaming, non-blocking fashion. But given that this is currently not possible I can only think of some kind of heuristics, where we could use buildozer to analyse some of the targets we get back from query (maybe even limiting it to filegroups), as well as the main target itself. Or maybe even simpler, if we have a target we just check if there is a filegroup of name like `<target>-globs` or `<target>-extra-watch-files`. None of these solutions are ideal but if well documented and/or opt-in then at least some people can get some file addition watching. What do you think?

Also just thinking about it in theory, what would be nice is to have something akin to the build event protocol / build event service, like loading event protocol, or analysis event protocol where bazel streams the information it finds during loading or analysis phase to a specified service with the nodes of the graph that it finds, and I suppose specifically for the watcher use-case also with some metadata about the original state of the input, i.e. this file is part of that glob.

Thinking about it further, if bazel query could give out some metadata about the original state of the rule attributes that would already help us get there and if ibazel had a flag to allow to set a different `--output_base` people who know that all of their external repositories are cached in the repository_cache could already make use of parallelization today.

I often like thinking about how could we get to a solution in the most ideal way and how can we actually get to a solution today, hence there are just a lot of thoughts spilled out and they evolved a bit as I was writing but curios to hear your thoughts on the topic.

pward123:
[...]

PR #180, opened 2018-11-27
Watch files via parent directory

### LINKED ISSUE #173: Second and further file changes not detected on MacOS Mojave
Possibly related to https://github.com/bazelbuild/bazel-watcher/pull/139 and https://github.com/angular/angular/issues/21770

I've tested this using bazel-watcher version 0.7.0 and 0.5.0, on MacOS Mojave.
As an example project I picked `java-tutorial` from https://github.com/bazelbuild/examples.git .
Checkout that repository, change into the `java-tutorial` directory and execute `ibazel build //:ProjectRunner`. Then open the files `src/main/java/com/example/{Greeting.java,ProjectRunner.java}` in an editor.
- Enter an empty line into the first file and save.
    `ibazel` notices and rebuilds the target.
- Enter a second empty line into the first file and save.
    `ibazel` does not notice the modification and does not rebuild the target.
    Any further edits to this file are not noticed.
- Enter an empty line into the second file and save.
    `ibazel` notices and rebuilds the target.
- Enter a second empty line into the second file and save.
    `ibazel` does not notice the modification and does not rebuild the target.
    Any further edits to this file are not noticed.

We have observed the issue on two separate MacOS Mojave machines. We have observed it with version `0.5.0` and `0.7.0`. We have not seen an instance of `ibazel` working properly on MacOS.

---

**Update**: The issue was observed on MacOS Sierra (10.12.6) and Mojave (10.14).
The issue was not observed on Linux, bazel-watcher seems to be working fine on Linux.

dougkoch:
Fixes #173

### DIFF
--- a/ibazel/ibazel.go
+++ b/ibazel/ibazel.go
@@ -71,7 +71,7 @@ type IBazel struct {
 	buildFileWatcher  *fsnotify.Watcher
 	sourceFileWatcher *fsnotify.Watcher
 
-	filesWatched map[*fsnotify.Watcher]map[string]bool // Inner map is a surrogate for a set
+	filesWatched map[*fsnotify.Watcher]map[string]struct{} // Inner map is a surrogate for a set
 
 	sourceEventHandler *SourceEventHandler
 	lifecycleListeners []Lifecycle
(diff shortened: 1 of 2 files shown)


Globegitter:
Ahh interesting idea - I have not looked at the code but watching the parent directory rather than the source files directly should also make it possible to watch these directories for newly added files right @achew22? That could be a simple workaround to fix #135

achew22:
@Globegitter, this will help with that problem. We will still have the problem that we don't know what files were included by glob and which were included directly by filenames. This will definitely help with the lack of triggering on Macs which is the primary focus of the change and should be celebrated for that.

Great work Doug! You rock, as always. :tada: 

PS: I think my favorite thing about this PR is that the e2e tests didn't have to change. Nicely done.

PR #188, opened 2018-12-03
Updates WORKSPACE file to work with bazel 0.20.0

Jdban:
git_repository and http_archive are deprecated in bazel version 0.20.0, this allows them to still work


achew22:
Passes locally

PR #195, opened 2018-12-05
Fix the output runner

borkaehw:
The problem is introduced by 'Uncross stdout and stderr buffers' https://github.com/bazelbuild/bazel-watcher/commit/add8041bfb843d5566e18b6711d88e4e10c5b699. Since we switched stdoutBuffer and stderrBuffer, `output_runner` is not reading the correct output log.


achew22:
Thanks for the investigation and the fix!

PR #198, opened 2018-12-06
Add e2e test to verify output runner functionality

### LINKED ISSUE #196: How do we test if output_runner is reading the output correctly
`output_runner` was introduced few months ago, which is a good feature. We recently rebased iBazel from master but `output_runner` is broken because of https://github.com/bazelbuild/bazel-watcher/commit/add8041bfb843d5566e18b6711d88e4e10c5b699. The test for `output_runner` is not sufficient since we are not testing if `output_runner` can correctly read the log from Bazel.

My question is what is the best way to run a Bazel command in Bazel test to simulate there is a dependency issue and `buildozer` command is generated? And we should run the `buildozer` command in Bazel test to see if `output_runner` can fix the `BUILD.bazel` correctly, if possible.

achew22:
Fixes #196

### DIFF
--- a/e2e/BUILD
+++ b/e2e/BUILD
@@ -11,6 +11,6 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "@com_github_bazelbuild_bazel_integration_testing//go:go_default_library",
-        "@com_github_bazelbuild_rules_go//go/tools/bazel:go_default_library",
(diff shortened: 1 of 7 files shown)


dougkoch:
@achew22 I'm fairly sure that the Travis CI build will succeed if you undo your shortening of the e2e timeout.

PR #200, opened 2018-12-06
Stay alive after source query failure

### LINKED ISSUE #194: ibazel should not crash if a build file change results in a syntax error
given `ibazel test //foo/bar`

if I make a typo in `foo/bar/BUILD.bazel`

ibazel process crashes and prints out the underlying bazel error

I expect the watcher to display the error, but keep on watching the same files as before. It's likely that the next file change will correct the build error and I'll be able to run the tests again.

The current behavior breaks the development flow and creates significant ergonomics issue for anyone using ibazel.

dougkoch:
Fixes #194


IgorMinar:
Thank you!

PR #231, opened 2019-03-11
Added newline to 'Starting...' text

zaucy:
I added a newline after the 'Starting...' text so that applications that depend on the output to start at position 0 are properly displayed while using `ibazel run`

**Before**
image

**After**
image


achew22:
Looks good to me. Thanks for sending in your first PR, @zaucy and welcome to the Bazel club! I'm probably not going to do a release just for this. Is that okay with you? If not, can you comment here and I'll do a minor release with this fix in it.

Thanks again!

zaucy:
@achew22 Yeah that's totally fine. It's such a small change. Thanks for the review

PR #253, opened 2019-05-01
Fix windows binary resolution on node

rerion:
Relevant issue:

https://github.com/bazelbuild/bazel-watcher/issues/105#issuecomment-487960158


achew22:
I just released this to NPM as `v0.10.2`. Could I have you confirm this is fixed now?

rerion:
Yes, this is fixed

PR #259, opened 2019-05-20
Update dependency on bazel-integration-testing

laurentlb:
It contains Bazel compatibility fixes (e.g.
incompatible_string_join_requires_strings)


achew22:
I just merged #254 which should have the same effect. Thanks!

Issue #248, opened 2019-04-09
Make livereload usable for custom rules 


Ubehebe:
First, thanks for your work on ibazel. It's an important part of my development workflow, as I discuss below.

I'd like to start a conversation about making ibazel's livereload feature easier to integrate with custom rules. Here's the current situation. If I have an executable Starlark rule foo, `ibazel run :some_foo_target` works OK without any modification to the foo rule implementation. When any of the target's inputs change, ibazel kills whatever the foo target was doing and restarts it with the fresh files. For executable rules that do something like print to a terminal, this is an ideal development experience.

For rules that do something more expensive/complex like bring up a web server and open a browser, the base ibazel behavior is not a good development experience. (For example, in a webapp rule I've written, editing a CSS or JS file causes ibazel to kill both the server and the browser, then bring them back a few seconds later.) It would be great if there was a simple API for custom rules to gain livereload superpowers.

My vision is this: **any rule that produces JS, CSS, or HTML (or any outputs that it is valuable to show in a browser) should be able to be invoked with ibazel run and have livereload just work**.

Currently, it is possible to get a custom rule to work with livereload, but it isn't easy. The rule has to set up a web server that can read the magic `IBAZEL_LIVERELOAD_URL` environment variable on startup and inject script tags pointing to that URL in the HTML files it serves. ts_devserver does this, but it also does a lot of other stuff (concatenating JS, injecting require.js) that may not be appropriate for custom rules.

I think there is an opportunity here to make ibazel useful for a great deal of frontend toolchains by providing this logic. I've implemented it in a private repository, and would be willing to do the work to upstream it here. But before talking about technical details, I wanted to get your overall thoughts.

Maintainers, what do you think? Does this fit in with the scope and purpose of this project? (I know there's some overlap between ibazel and ts_devserver, and they seem to have been developed tightly coupled. I'm asking here first because the ibazel repo seems like the right level of generality for this logic.)

achew22:
Thanks for the kind words!

> Maintainers, what do you think? Does this fit in with the scope and purpose of this project? (I know there's some overlap between ibazel and ts_devserver, and they seem to have been developed tightly coupled. I'm asking here first because the ibazel repo seems like the right level of generality for this logic.)

[...]

PR #261, opened 2019-05-24
Merge bazel-runfiles-server into upstream

Ubehebe:
This PR merges my [bazel-runfiles-server](http://github.com/Ubehebe/bazel-runfiles-server) repo into bazel-watcher. See https://github.com/bazelbuild/bazel-watcher/issues/248 for motivation and further discussion.

Maintainers, how can I help you review this? I've kept the complete commit history from the downstream repo, but can squash if you prefer. I've tried to follow the existing style in both the golang and Starlark sources.


Ubehebe:
Looks like the Windows tests are failing. The integration test runner waits for the system under test to print a line to its stdout before running the test binary. The line delimiter is [hardcoded](https://github.com/Ubehebe/bazel-watcher/blob/runfiles-server-pr/runfiles_server/integrationtest/integration_test_runner.go#L54) to be `\n`; it probably needs to be `\r` on Windows.

But that's just the proximate cause. The integration test binaries themselves are written in bash and use curl, so I imagine there are plenty of problems ahead. What do you think is the way forward? Ideas:

- Port the tests to Go
- Write .cmd test binaries for Windows
- Disable the Windows tests for this PR and get them working later

achew22:
@Ubehebe, looking through the tests I think my inclination would be to go for the first option, "Port the tests to Go". It looks to me like there are only about 4 tests and they aren't extremely complex. I'd rather push the multi-os compatibility into a language designed for it rather than having 2 or 3 implementations or each test.

Ubehebe:
SGTM. I will be AFK for a few days, but should be able to port the tests by the end of the week.

paynecodes:
@Ubehebe any tips for using this approach today? Other than maintain my own version of ibazel that is :)

PR #262, opened 2019-05-24
Add stopAfter to output_runner options

### LINKED ISSUE #260: Give output runner an option to exit early
Sometimes the build will output 100+ errors but the fix for the first will fix all 100+ of them.
Currently with output_runner, the fix will be applied from the first error, and then for all of the other errors the tool may try to fix the issue, but the issue is likely already resolved from the first fix.  Not only can this cause issues with certain tools, but can also cause massively increase the time to complete a build.

I think there are two reasonable solutions to this:
1) If the "command" specified in `.bazel_fix_commands.json` returns a specific error code then immediately dump any existing output and restart the build.
2) Add an option to `.bazel_fix_commands.json` that specifies that a match should only be applied once per build.

DavidANeil:
Resolves #260.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,30 @@ your target contains `ibazel_notify_changes` in its `tags` attribute, then the
 command will stay alive and will receive a notification of the source changes on
 stdin.
 
+## Output Runner
+
+iBazel is capable of producing and running commands from the output of Bazel commands. If iBazel is run with the flag `--run_output` then it will check for a `%WORKSPACE%/.bazel_fix_commands.json` and if present run any commands that match the provided regular expressions.
+For example the commands defined by the following file will match `buildozer` commands found in the output and provide a prompt to run the command.
+```
+[{
+    "regex": "^buildozer '(.*)'\\s+(.*)$",
+    "command": "buildozer",
+    "args":    ["$1", "$2"],
+    "stopAfter": 127,
+}]
+```
+Adding the flag `--run_output_interactive=false` will automatically run the command without prompting for confirmation.
+The fields in `.bazel_fix_commands.json` are:
+
+* regex: a regular expression that will be matched against every line of output.
+    * backslash `\` characters will need to be escaped once for the regex to be parsed properly.
+* command: a command that will be run from the workspace root.
+* args: a list of arguments to provide to the command, with `$1` being the first match group of `regex`, `$2` being the second and so on.
+* stopAfter (optional): a number (0-127) of failures (non-0 exit codes from the `command`) after which to discard any remaining output.
+    * default value: 127
+    * no-op with `--run_output_interactive=true`
+
+
 ## Profiling
 
 iBazel has a `--profile_dev` flag which turns on a generated profile output file
@@ -123,7 +147,7 @@ The profile outfile is in concatenated JSON format. Events are outputted one per
 You can find an example profile output JSON file [here](https://github.com/bazelbuild/bazel-watcher/blob/master/example.profile.json). Below is the file in pretty print JSON format:
 
 ```
-{  
+{
    "type":"IBAZEL_START",
    "iteration":"4214114686684e0f",
    "time":1513706108351,
@@ -132,35 +156,35 @@ You can find an example profile output JSON file [here](https://github.com/bazel
    "maxHeapSize":"3817MB",
    "committedHeapSize":"1372MB"
 }
-{  
+{
    "type":"RUN_START",
    "iteration":"4214114686684e0f",
    "time":1513706109329,
    "targets":["//src:devserver"],
    "elapsed":978
 }
-{  
+{
    "type":"RELOAD_TRIGGERED",
    "iteration":"4214114686684e0f",
    "time":1513706114595,
    "targets":["//src:devserver"],
    "elapsed":6244
 }
-{  
+{
    "type":"RUN_DONE",
    "iteration":"4214114686684e0f",
    "time":1513706114595,
    "targets":["//src:devserver"],
    "elapsed":6244
 }
-{  
+{
    "type":"SOURCE_CHANGE",
    "iteration":"7e6f8e150e9a8367",
    "time":1513706129384,
    "targets":["//src:devserver"],
    "change":"/Users/greg/google/gregmagolan/angular-bazel-example/src/hello-world/hello-world.component.ts"
 }
-{  
+{
    "type":"RUN_START",
    "iteration":"7e6f8e150e9a8367",
    "time":1513706129484,
@@ -168,7 +192,7 @@ You can find an example profile output JSON file [here](https://github.com/bazel
    "elapsed":100,
    "changes":["/Users/greg/google/gregmagolan/angular-bazel-example/src/hello-world/hello-world.component.ts"]
 }
-{  
+{
    "type":"RELOAD_TRIGGERED",
    "iteration":"7e6f8e150e9a8367",
    "time":1513706133947,
@@ -176,7 +200,7 @@ You can find an example profile output JSON file [here](https://github.com/bazel
    "elapsed":4563,
    "changes":["/Users/greg/google/gregmagolan/angular-bazel-example/src/hello-world/hello-world.component.ts"]
 }
-{  
+{
    "type":"RUN_DONE",
    "iteration":"7e6f8e150e9a8367",
    "time":1513706133947,
@@ -184,7 +208,7 @@ You can find an example profile output JSON file [here](https://github.com/bazel
    "elapsed":4563,
    "changes":["/Users/greg/google/gregmagolan/angular-bazel-example/src/hello-world/hello-world.component.ts"]
 }
-{  
+{
    "type":"REMOTE_EVENT",
    "iteration":"7e6f8e150e9a8367",
    "time":1513706134297,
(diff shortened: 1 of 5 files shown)


achew22:
Hey @DavidANeil, thanks so much for the PR! I'd really like to merge this. Two things that I would like to see before I do. One is a blocker and one is a nice to have.

Blockers:

* [ ] Add some tests. There is a nice [output_runner_test](https://github.com/bazelbuild/bazel-watcher/blob/master/e2e/output_runner/output_runner_test.go) you can expand from but I think it has the skeleton of what we need.

Nice to have:

* [ ] Could I have you add a section to `README.md` to explain how to use this feature?

Thanks so much!

DavidANeil:
Thank you for the feedback, I have added a test and added documentation for the output runner.
Let me know if there are any nitpicks you have: I am happy to get this right.

README.md
```diff
@@ -67,6 +67,30 @@ your target contains `ibazel_notify_changes` in its `tags` attribute, then the
 command will stay alive and will receive a notification of the source changes on
 stdin.
 
+## Output Runner
```
achew22: Thanks so much for adding this! This is really great documentation.

README.md
```diff
@@ -67,6 +67,30 @@ your target contains `ibazel_notify_changes` in its `tags` attribute, then the
...
+}]
+```
+Adding the flag `--run_output_interactive=false` will automatically run the command without prompting for confirmation.
+The fields in `.bazel_fix_commands.json` are:
+
+* regex: a regular expression that will be matched against every line of output.
+    * backslash `\` characters will need to be escaped once for the regex to be parsed properly.
+* command: a command that will be run from the workspace root.
+* args: a list of arguments to provide to the command, with `$1` being the first match group of `regex`, `$2` being the second and so on.
+* stopAfter (optional): a number (0-127) of failures (non-0 exit codes from the `command`) after which to discard any remaining output.
```
achew22: WRT `stopAfter` I'm trying to think of the way people are going to experience this. To my naive eye there are 2 modes that these generator tools operate. First you have the tools that emit a single build instruction like `gazelle` who rewrite all the `BUILD` files in the repo (that it knows how to write). You would configure something that invokes Gazelle.

```
{
    "regex": "^Please run gazelle$",
    "command": "bazel",
    "args":    ["run", "--", "-mode=fix"],
    "stopAfter": 1, // <-- This is 1 because running Gazelle multiple times will never help, so coalesce all the calls into a single invocation.
}
```

Where something like `buildozer` might have bunch of events that were emitted and you would configure it this way:

```
{
    "regex": "^buildozer '(.*)'\\s+(.*)$",
    "command": "buildozer",
    "args":    ["$1", "$2"],
    "stopAfter": 127, // <-- It will invoke this up to 127 times.
}
```

What isn't clear to me is how it is determining what to run up to 127 times. I could interpret this to mean that it will invoke `buildozer add dep //blah //foo` 127 times until it gets an exit status of `0`, or that it is going to put all the `buildozer` commands into a queue and run `127` of them in a row before giving up?

DavidANeil: I think there are 3 types of scenarios.
1 is the `gazelle` type fix. `stopAfter` is not helpful here.
2 is the `rules_scala_annex` type fix where it detects problems directly with the build file and outputs one error.  This leads to 1 error to 1 action mapping.  `stopAfter` is not helpful here.
3 is the `rules_typescript` type fix where the compiler identifies a missing a dependency, but then outputs an error for for each import of each file that is missing a dependency.  For the code base I am working on that involved about 3000 errors that mapped to 27 actions.  `stopAfter` is designed for this case.  The idea is that it will fix the first 3-4 errors, then try the compile step again, and hopefully only get 1000 errors, then do a few more actions, and so on until the compiler does not output any more things for the output_runner to handle.

In my case I use `"stopAfter": 3` to allow for some malformed commands, but to generally stop and restart as soon as possible.

So your second clause is the correct one, it will not try to parse the same output more than once, but if different output is matched by the same command regex, then it will only run it until it has failed `stopAfter` number of times. (It will in fact stop running all commands after 1 exceeds the `stopAfter` limit.)
If you have a way to phrase this in the documentation to make it more clear, let me know.

DavidANeil:
> To expand on this a little further, I have 2 relatively high level questions. I think the code counts 1 per failure/runner and after 127 failures it stops executing more, why is this a per-runner configuration option? Should this be done as a `ibazel` level flag?

I did this so that one workspace could have different tolerances for different outputs.  Some tools are better than others at giving great output and I figured this would be the most robust solution.  I originally wanted to specify a list of error codes on which to terminate immediately, but I couldn't find a great cross-platform way of doing that.

> Also, whenever I have these commands run on my projects it inevitably changes a `BUILD` file (duh, I know) so don't you still end up in this loop because even if one of the `buildozer` command fails it's still going to rerun the test and then it'll emit the same buggy output but with whichever entry succeeded removed. So, right now, `ibazel` will converge on only the set of `buildozer` operations that fail and will keep looping. In the event that you set this number too low, let's say 1 just as an example, then you could find yourself in a situation where the first `buildozer` run crashes and then you rerun the test, get the same output and loop. Am I mistaken in that?

You could put yourself in a state where the output_runner will not be helpful by setting it too low, but because the failed command will have also not changed the `BUILD` file, after output_runner is done `ibazel` will simply wait for a dependency to change before it will run again.  This should prevent infinite loops in most cases.

DavidANeil:
> To expand on this a little further, I have 2 relatively high level questions. I think the code counts 1 per failure/runner and after 127 failures it stops executing more, why is this a per-runner configuration option? Should this be done as a `ibazel` level flag?

I did it this way so that one workspace could have different tolerances for different outputs.  Some tools are better than others at giving great output.

> Also, whenever I have these commands run on my projects it inevitably changes a `BUILD` file (duh, I know) so don't you still end up in this loop because even if one of the `buildozer` command fails it's still going to rerun the test and then it'll emit the same buggy output but with whichever entry succeeded removed. So, right now, `ibazel` will converge on only the set of `buildozer` operations that fail and will keep looping. In the event that you set this number too low, let's say 1 just as an example, then you could find yourself in a situation where the first `buildozer` run crashes and then you rerun the test, get the same output and loop. Am I mistaken in that?

If you set it to 1 and the command files without modifying the build file then nothing will have changed and no rebuild will be triggered.

DavidANeil:
I noticed while testing that the existing `TestOutputRunner` had some flakiness issues, and it appears the `TestStopAfter` suffers from the same issue.  Do you have any recommendations for stabilizing it?

achew22:
> You could put yourself in a state where the output_runner will not be helpful by setting it too low, but because the failed command will have also not changed the BUILD file, after output_runner is done ibazel will simply wait for a dependency to change before it will run again. This should prevent infinite loops in most cases.

Good point, if it is broken it won't update the `BUILD` file. Given that constraint, what would you think about just setting it to 100 and not having it as a knob? What are the circumstances where someone would want to set it to more/less than that? Can you help me understand the value in making it configurable?

> I noticed while testing that the existing TestOutputRunner had some flakiness issues, and it appears the TestStopAfter suffers from the same issue. Do you have any recommendations for stabilizing it?

Do you know in what way it is flaking?

DavidANeil:
For us, fixing the BUILD file involves running 2 `bazel query`s, which can take up to a second, so running all of the commands took about 12 hours without this change (and I was occasionally simulating this change by manually killing ibazel and starting over after I saw lots of errors in a row), and about 1 hour with `stopAfter: 3` with no intervention.

The flakes are that the command is either not running or the file is not being created under the expected name.
If it matches what I have seen, there are probably backspace (or other curses) characters in the output which prevents them from being matched.

DavidANeil:
I have tracked it down the the 5ms sleep in `Expect` in `e2e/ibazel.go`.  If the output_runner takes more than 5ms to run then the commands/output_runner/lifecycle will be terminated.   This would occasionally happen with `TestOutputRunner` but happens very often with my `TestStopAfter`.  I can bump the interval, but this is definitely a race condition.  This project is my first experience with Go, so I would love pointers on how to fix this properly.
I am going to push a commit with the bumped interval because that is the only way I know to get buildkite to build it.

DavidANeil:
I have tracked it down further to what is happening in these tests:
1) Build action is triggered.
2) Polling begins waiting for "action" on stdout.
2) Output that matches output_runner will be sent on stderr.
3) The action completes. "action" is sent to stdout.
4) Lifecycle hook is triggered, output_runner begins running.
Sometimes) output_runner completes running.
5) poll ticks and sees "action" in stdout.
6) Test checks if output_runner has successfully run.
7) Test ends.  output_runner and its sub-processes are terminated.
Never) output_runner completes running.

Would it make sense for there to be some output when all lifecycles have completed?  That seems excessive for allowing for testing, so I'll let you make the decision on this one.

achew22:
Thanks for such thorough investigation! To ask a quick follow up

> Would it make sense for there to be some output when all lifecycles have completed? That seems excessive for allowing for testing, so I'll let you make the decision on this one.

How do you imagine that event being uttered? Would it be sent only when output_runners find something?

achew22:
Also, wrt increasing the timeout I think that's probably a reasonable course of action. `ibazel` is no being invoked in go but instead as a subprocess. 5ms seems like an awful small amount of time. I think 500ms sounds reasonable

DavidANeil:
This was a bandage for a different problem, going to close this.
Would you like me to open a PR with the documentation though?

achew22:
@DavidANeil, that would be really great if you could do that, thanks!

PR #265, opened 2019-05-29
Notify listening processes when a build starts.

DavidANeil:


achew22:
@DavidANeil, looks good to me. Merging.

PS: What are you going to use this for? It seems like you're putting some work into iBazel lately. Any chance it's open source? I'd love to see how things are being used.

DavidANeil:
We have been using Bazel for our backend with iBazel for developers and now we are getting frontend Bazel set up and so all of my PRs relate to helping the frontend experience be great for developers.  Which is why there has been an uptick recently.

This PR is to allow our asset server to just be running an iBazel build on all of the transpiled sources and then serve them immediately if a build is not in progress, rather than waiting for a request for an asset, then building it, then serving it.

https://github.com/bazelbuild/bazel-watcher/pull/262 is to automatically add missing dependencies to BUILD files for our TypeScript.

https://github.com/bazelbuild/bazel/pull/8322 is to make it so we can use iBazel to debug our Karma/Jasmine unit tests.  So the chrome window can stay open.

Unfortunately none of our use of this is in open sourced code.

PR #271, opened 2019-07-09
Pass bazel args for run command.

### LINKED ISSUE #269: run command should also forward recognized flags
As mentioned in the <https://github.com/bazelbuild/bazel-watcher/issues/242#issuecomment-478871815> run does not forward `overrideableBazelFlags` to the `bazel run` command.

https://github.com/bazelbuild/bazel-watcher/blob/18bdb44832ccc533e0ab3923ef80060eeb24582d/bazel/bazel.go#L210

This can be really useful for running watcher on binaries in dev mode like

```sh
ibazel run --config=dev //cmd/server

### PR DESCRIPTION
**Pull request by zoido**
Fixes #269

----
also:
 - extends `IBazelTester` so that Bazel flags can be passed
 - adds an e2e test covering the case

### DIFF
--- a/e2e/ibazel.go
+++ b/e2e/ibazel.go
@@ -44,18 +44,18 @@ func (i *IBazelTester) Build(target string) {
 }
 
 func (i *IBazelTester) Run(target string, bazelArgs ...string) {
-	i.run(target, []string{}, bazelArgs)
+	i.run(target, bazelArgs, []string{})
 }
 
 func (i *IBazelTester) RunWithProfiler(target string, profiler string) {
-	i.run(target, []string{"--profile_dev=" + profiler}, []string{})
+	i.run(target, []string{}, []string{"--profile_dev=" + profiler})
 }
 
 func (i *IBazelTester) RunWithBazelFixCommands(target string) {
-	i.run(target, []string{
+	i.run(target, []string{}, []string{
 		"--run_output=true",
 		"--run_output_interactive=false",
-	}, []string{})
+	})
 }
 
 func (i *IBazelTester) GetOutput() string {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: e2e/ibazel.go [hunk 1]
```diff
@@ -157,11 +157,12 @@ func (i *IBazelTester) build(target string, additionalArgs []string) {
 	}
 }
 
-func (i *IBazelTester) run(target string, additionalArgs []string) {
+func (i *IBazelTester) run(target string, additionalArgs []string, bazelArgs []string) {
```
achew22: Could you move the bazelArgs to be the 2nd parameter to match the order that they are passed into Bazel?

zoido: Done :heavy_check_mark:. PTAL.

e2e/ibazel.go
```diff
@@ -43,16 +43,16 @@ func (i *IBazelTester) Build(target string) {
 	i.build(target, []string{})
 }
 
-func (i *IBazelTester) Run(target string) {
-	i.run(target, []string{})
+func (i *IBazelTester) Run(target string, bazelArgs ...string) {
```
achew22: I'm terribly sorry, but could you also flip the argument ordering here. It should always be `bazelArgs, target, [targetArgs]`

zoido: No problem. However then I see possible inconsistency with `.run()` that goes like `(target, bazelArgs, ibazelArgs)`

What pattern am I missing? :)

Also `.Run()` has it (at least for now) as _optional_.

zoido: DONE. ✔️ PTAL.

achew22: You're 100% right that [`.run`](https://github.com/bazelbuild/bazel-watcher/blob/9a15dd18d8312b3b2770afd27564009296bf25fb/e2e/ibazel.go#L160) should be flipped around. Maybe it should even be a dictionary that maps from flag to value, but the reason to be persnickety about this is because it is part of the public interface. `.run` is internal to this function and I can fix it later without breaking anyone.

Thanks again for fixing this

PR #275, opened 2019-07-16
feat: locate Bazel binary in sibling package when installed via npm

alexeagle:
- probably needs some added unit testing
- should verify it works on Windows

at a higher level, this raises the question of why `@bazel/bazel` and `@bazel/ibazel` are distributed as two different packages. I suppose it's because they are on independent release cycles, but we could probably decide to line those up somehow? It would be slightly easier to locate the other binary if they were distributed together.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -23,13 +23,66 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	blaze_query "github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf"
 	"github.com/golang/protobuf/proto"
 )
 
-var bazelPath = flag.String("bazel_path", "bazel", "Path to the bazel binary to use for actions")
+var bazelPathFlag = flag.String("bazel_path", "", "Path to the bazel binary to use for actions")
+
+// bazelNpmPath looks up a relative path to a binary from @bazel/bazel
+// This is used as an alternate resolution when no bazel binary is in the $PATH
+// When running from the @bazel/ibazel npm package, our binary is
+// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+// We can find bazel in
+// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+func bazelNpmPath(ibazelBinPath string) (string, error) {
+	s := strings.Split(ibazelBinPath, "/")
+	for i := 0; i + 4 < len(s); i++ {
+		nm, scope, pkg, dir, bin := s[i], s[i+1], s[i+2], s[i+3], s[i+4]
+		if nm == "node_modules" && scope == "@bazel" && pkg == "ibazel" && dir == "bin" {
+			// See mapping in release/npm/index.js - ibazel is named with "amd64" arch 
+			// but @bazel/bazel uses node arch names
+			arch := strings.Replace(strings.Replace(bin, "amd64", "x64", 1), "-", "_", 1)
+			dir := strings.Join(append(s[0:i+2], "bazel-" + arch), "/")
+			// Find the bazel binary in the directory - it will have a version number in the name
+			// so we list all the files and find a bazel-*-$ARCH
+			if fd, err := os.Open(filepath.FromSlash(dir)); err == nil {
+				if names, err := fd.Readdirnames(0); err == nil {
+					for j := 0; j < len(names); j++ {
+						if strings.HasPrefix(names[j], "bazel-") {
+							return dir + "/" + names[j], nil
+						}
+					}
+				}
+			}
+		}
+	}
+	return "", errors.New("bazel binary not found in @bazel/bazel package")
+}
+
+func findBazel() (string) {
+	// Trust the user, if they supplied a path we always use it
+	if len(*bazelPathFlag) > 0 {
+		return *bazelPathFlag
+	}
+	// Common case, check in $PATH for system-installed Bazel
+	if path, err := exec.LookPath("bazel"); err == nil {
+		return path
+	}
+	// Frontend devs may have installed @bazel/bazel and @bazel/ibazel from npm
+	if npmPath, err := bazelNpmPath(filepath.ToSlash(os.Args[0])); err == nil {
+		return filepath.FromSlash(npmPath)
+	}
+
+	// If we've fallen through to here, the lookup won't succeed.
+	// Return "bazel" so that we'll later fail with an error
+	//   exec: "bazel": executable file not found in $PATH
+	// which helps the user understand that we looked in the $PATH
+	return "bazel"
+}
 
 type Bazel interface {
 	SetArguments([]string)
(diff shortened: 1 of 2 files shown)


bazel/bazel.go
```diff
@@ -23,13 +23,58 @@ import (
...
+
+		// No bazel binary is in the $PATH
+		// When running from the @bazel/ibazel npm package, our binary is
+		// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+		// We can find bazel in
+		// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+		fmt.Printf("didn't find bazel in the $PATH, %v\n", os.Args[0])
+		s := strings.Split(filepath.ToSlash(os.Args[0]), "/")
+		for i := 0; i < len(s); i++ {
+			fmt.Printf("segment %v\n", s[i])
```
achew22: Can you break the segments down into meaningful names? I have a difficult time reasoning about s[i+2] vs s[i+3]

https://play.golang.org/p/cwPaZv7klcC

bazel/bazel.go
```diff
@@ -23,13 +23,58 @@ import (
...
+		}
+
+		// No bazel binary is in the $PATH
+		// When running from the @bazel/ibazel npm package, our binary is
+		// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+		// We can find bazel in
+		// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+		fmt.Printf("didn't find bazel in the $PATH, %v\n", os.Args[0])
+		s := strings.Split(filepath.ToSlash(os.Args[0]), "/")
+		for i := 0; i < len(s); i++ {
```
achew22: Could you break this logic into a function that takes a `[]string` and returns `path string, err error` and add a unit test

bazel/bazel.go
```diff
@@ -23,13 +23,58 @@ import (
...
+			return path
+		}
+
+		// No bazel binary is in the $PATH
+		// When running from the @bazel/ibazel npm package, our binary is
+		// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+		// We can find bazel in
+		// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+		fmt.Printf("didn't find bazel in the $PATH, %v\n", os.Args[0])
+		s := strings.Split(filepath.ToSlash(os.Args[0]), "/")
```
achew22: I'm curious if you have thoughts on how to write a test for the stuff north of here? I guess you could make a synthetic FS and set the path to contain it, but that feels not great.

bazel/bazel.go
alexeagle: I'm pretty novice at go so I don't know if there's a way to go straight from a slice like
`a,b,c,d := s[i:i+4]...`

achew22: I think you can do it this way: https://play.golang.org/p/1f15O3khaB2

bazel/bazel.go
alexeagle: Done

bazel/bazel.go
alexeagle: I refactored it a bit so the conditional logic is brain-dead simple. I think this means we just don't try to test these two conditions.

bazel/bazel_test.go
```diff
@@ -91,3 +91,21 @@ func TestCancel(t *testing.T) {
 	b := New()
 	b.Cancel()
 }
+
+func TestBazelNpmPath(t *testing.T) {
+	// Where ibazel gets installed by npm
+	ibazelPath := os.Getenv("TEST_TMPDIR") + "/node_modules/@bazel/ibazel/bin/linux-amd64/ibazel"
```
achew22: Can you turn this into a [table driven test](https://github.com/golang/go/wiki/TableDrivenTests), and add a few negative test cases?

bazel/bazel.go
```diff
@@ -23,13 +23,66 @@ import (
...
+// This is used as an alternate resolution when no bazel binary is in the $PATH
+// When running from the @bazel/ibazel npm package, our binary is
+// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+// We can find bazel in
+// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+func bazelNpmPath(ibazelBinPath string) (string, error) {
+	s := strings.Split(ibazelBinPath, "/")
+	for i := 0; i + 4 < len(s); i++ {
+		nm, scope, pkg, dir, bin := s[i], s[i+1], s[i+2], s[i+3], s[i+4]
+		if nm == "node_modules" && scope == "@bazel" && pkg == "ibazel" && dir == "bin" {
```
achew22: Thanks for doing this! The conditional here is, indeed, very simple to reason through.

bazel/bazel.go
```diff
@@ -23,13 +23,66 @@ import (
...
+
+// bazelNpmPath looks up a relative path to a binary from @bazel/bazel
+// This is used as an alternate resolution when no bazel binary is in the $PATH
+// When running from the @bazel/ibazel npm package, our binary is
+// /DIR/node_modules/@bazel/ibazel/bin/$ARCH/ibazel
+// We can find bazel in
+// /DIR/node_modules/@bazel/bazel-$ARCH/bazel-0.28.0-linux-x86_64
+func bazelNpmPath(ibazelBinPath string) (string, error) {
+	s := strings.Split(ibazelBinPath, "/")
+	for i := 0; i + 4 < len(s); i++ {
```
achew22: No action required, but since you know a lot about compilers you might have a neat trick here. The two are very definitely equivalent, Why `i + 4 < len(s)` instead of `i < len(s) - 4`?

Do compilers do anything where they detect that the inside of the loop isn't modifying either side of the operation, precomputing the rhs of the inequality and then having a const `JNLE` instruction? You probably don't know anything about the GoLang compiler in particular, but I'm curious none the less.

bazel/bazel.go
```diff
@@ -23,13 +23,66 @@ import (
...
+		nm, scope, pkg, dir, bin := s[i], s[i+1], s[i+2], s[i+3], s[i+4]
```
achew22: Could you extract a "prefix" as 

		prefix, nm, scope, pkg, dir, bin := s[0:i], s[i], s[i+1], s[i+2], s[i+3], s[i+4]

bazel/bazel.go
```diff
@@ -23,13 +23,66 @@ import (
...
+		if nm == "node_modules" && scope == "@bazel" && pkg == "ibazel" && dir == "bin" {
+			// See mapping in release/npm/index.js - ibazel is named with "amd64" arch 
+			// but @bazel/bazel uses node arch names
+			arch := strings.Replace(strings.Replace(bin, "amd64", "x64", 1), "-", "_", 1)
+			dir := strings.Join(append(s[0:i+2], "bazel-" + arch), "/")
```
achew22: Here can you use `append(prefix, nm, scope, package, "bazel-"+arch)`

bazel/bazel.go
```diff
@@ -23,13 +23,66 @@ import (
...
+	// Trust the user, if they supplied a path we always use it
+	if len(*bazelPathFlag) > 0 {
+		return *bazelPathFlag
+	}
+	// Common case, check in $PATH for system-installed Bazel
+	if path, err := exec.LookPath("bazel"); err == nil {
+		return path
+	}
+	// Frontend devs may have installed @bazel/bazel and @bazel/ibazel from npm
+	if npmPath, err := bazelNpmPath(filepath.ToSlash(os.Args[0])); err == nil {
```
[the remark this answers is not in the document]
alexeagle: yeah I agree, changed

bazel/bazel.go
alexeagle: I just did it this way since the expression `i+4` appears below, I know that's the index I want to exist, so this way the expression in the loop condition is identical

Issue #278, opened 2019-07-31
Add support for bazelrc flag


libsamek:
Since there are only a few Bazel flags supported, it would be really useful to add support for `--bazelrc` to use for additional configuration.

`--config` is useful in this case, but it is only limited to `.bazelrc`.

achew22:
It's very simple to add support for a new flag, [right here](https://github.com/bazelbuild/bazel-watcher/blob/4259108b33a9318ba5d0b3edede162edf9f71fc8/ibazel/main.go#L27). Would you be willing to send in a PR to support this flag, and any others you can think of?

libsamek:
Yeah, I would. So I understand you correctly I just need to add it to `overrideableBazelFlags`?

achew22:
Yeah, that's exactly what you would need to do. Feel free to add any other flags you are interested in having through ibazel

libsamek:
This is resolved with #13316.

PR #280, opened 2019-08-01
Add support for startup option bazelrc

libsamek:
As discussed in #278 adding support for `--bazelrc` option. Since this is a startup option, additional changes were needed.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -102,10 +102,13 @@ type Bazel interface {
 
 type bazel struct {
 	cmd           *exec.Cmd
+
 	args          []string
+	startupArgs   []string
+
 	ctx           context.Context
 	cancel        context.CancelFunc
-	startupArgs   []string
+
 	writeToStderr bool
 	writeToStdout bool
 }
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -88,6 +88,7 @@ func findBazel() (string) {
 
 type Bazel interface {
 	SetArguments([]string)
+	SetStartupArgs([]string)
 	WriteToStderr(v bool)
 	WriteToStdout(v bool)
 	Info() (map[string]string, error)
@@ -104,6 +105,7 @@ type bazel struct {
 	args          []string
 	ctx           context.Context
 	cancel        context.CancelFunc
+	startupArgs   []string
 	writeToStderr bool
 	writeToStdout bool
 }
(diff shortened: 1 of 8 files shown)


ibazel/main.go
```diff
@@ -26,6 +26,8 @@ var Version = "Development"
 
 var overrideableBazelFlags []string = []string{
 	"--action_env",
+	"--announce_rc",
+	"--bazelrc",
```
achew22: Could you make a separate list for start-up flags?

libsamek:
@googlebot I signed it!

ibazel/main.go
libsamek: Yes, np.

libsamek: Added a new array for startup flags.

ibazel/main.go
```diff
@@ -57,7 +62,7 @@ target, run, build, or test the specified targets.
 
 Usage:
 
-ibazel [flags] build|test|run targets...
+ibazel build|test|run [flags] targets...
```
achew22: could you make this be 

```
ibazel [startup_flags] build|test|run [flags] targets...
```

unless I'm incorrect in thinking it'll work like that

bazel/bazel.go
```diff
@@ -104,6 +105,7 @@ type bazel struct {
 	args          []string
 	ctx           context.Context
 	cancel        context.CancelFunc
+	startupArgs   []string
```
achew22: Nit: coudl you group args and startupArgs together then put a newline after

```
args        []string
startupArgs []string

ctx    context.context
cancel context.CancelFunc
```

ibazel/main.go
libsamek: Yes, I expected that to work as well, however:
```
$ ./ibazel --bazelrc=bazelrc.test build //...
flag provided but not defined: -bazelrc
iBazel - Version v0.10.3-dirty

A file watcher for Bazel. Whenever a source file used in a specified
target, run, build, or test the specified targets.

Usage:

ibazel [flags] build|test|run targets...

Example:
...
```

I suspect it has to do something how `flags.Parse()` works.

Any suggestions?

bazel/bazel.go
libsamek: Fixed.

ibazel/main.go
achew22: I think aiming to keep the same CLI invocation is noble, but it looks like I need to revisit the logic there to make it actually behave that way. Could I have you file an issue requesting this and I'll add it to my backlog.

Thanks!

achew22:
@libsamek, thank you so much for taking the time to fix this and to enhance the project!

ibazel/main.go
libsamek: I just open an issue regarding this: #283.

PR #282, opened 2019-08-16
Set IBAZEL=true in process env

statik:
Thank you for making this tool!

My proposed change will allow programs to detect they are being run under
bazel-watcher.

I'd like to have this because I use a bazel wrapper script that automatically runs gazelle and does a bunch of other bazel tuning, and I want to do different things when being invoked by ibazel.

I have signed the Google Individual CLA on September 04, 2008.

### DIFF
--- a/e2e/simple/simple_test.go
+++ b/e2e/simple/simple_test.go
@@ -93,6 +93,27 @@ sh_binary(
 	ibazel.ExpectOutput("Started!")
 }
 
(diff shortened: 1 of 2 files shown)


achew22:
Could I have you add an e2e test that fails if this value isn't set? You can use the [simple](https://github.com/bazelbuild/bazel-watcher/tree/master/e2e/simple) test case as an example.

statik:
@achew22 thanks for considering this! I added the test case and confirmed that it fails without this change.

PR #299, opened 2019-11-08
Remove ANSI codes from output before matching

### LINKED ISSUE #263: Color output in console can conflict with output_runner regex
It would probably be best to have output_runner strip any color codes from the output before trying to match the regex.

DavidANeil:
Resolves #263


achew22:
@DavidANeil, thank you so much for taking the time to fix this!

PR #302, opened 2019-11-13
Add a logging module

achew22:
This logging module improves the quality of output, and makes using
ibazel easier.


ibazel/ibazel.go
```diff
@@ -127,33 +128,37 @@ func (i *IBazel) handleSignals() {
...
+			fmt.Fprint(os.Stderr, "\n")
+			log.Log("Subprocess killed from getting SIGTERM\n")
 			i.cmd.Terminate()
 		}
 		osExit(3)
 		return
 	case syscall.SIGHUP:
 		if i.cmd != nil && i.cmd.IsSubprocessRunning() {
-			fmt.Fprintf(os.Stderr, "\nSubprocess killed from getting SIGHUP\n")
+			fmt.Fprint(os.Stderr, "\n")
```
mrmeku: It feels a bit weird to have to fallback to the old mechanism of logging whenever you want to prefix your message with a newline. Maybe you could just add a helper method to the logger?

`log.NewLine()` perhaps?

ibazel/log/log.go
```diff
@@ -0,0 +1,62 @@
...
+func log(c color, msg string, args ...interface{}) {
+	fmt.Fprintf(writer, "%siBazel [%s]%s: ",
+		c,
+		timeNow().Format(time.Kitchen),
+		reset)
+	fmt.Fprintf(writer, msg, args...)
+	fmt.Fprintf(writer, "\n")
+}
+
+// Error implements Logger.Error.
```
mrmeku: Just for my own education is it idiomatic to have comments like this in go? They don't feel like they add value to me so I was wondering why you chose to include them

ibazel/ibazel.go
mrmeku: Alternatively you could have each logger method take an optional integer argument of how many newlines to prefix a message with.

ibazel/log/log.go
achew22: I won't say idiomatic cause I'm a neophyte, but it is a thing I regularly see in code from people who are much better at this than me. In most languages you declare the type you're implementing, so I think this is just to hint to a user that you're actually implementing that interface.

ibazel/ibazel.go
achew22: I think you're on to something here. Log should append a \n at the end of the line on its own which will un-clutter most of this.

e2e/ibazel.go
```diff
@@ -21,12 +22,11 @@ const (
 	defaultDelay = 20 * time.Second
 )
 
-var (
-	ibazelLogFile = filepath.Join(os.TempDir(), "ibazel_output.log")
-)
+var ()
```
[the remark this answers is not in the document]
achew22: Done

PR #318, opened 2019-12-23
Remove tap-bin from Home-brew formula.

BooneJS:
Following the directions for Mac (Homebrew)
$ brew tap bazelbuild/tap
Updating Homebrew...
$ brew tap-pin bazelbuild/tap
Error: Calling brew tap-pin user/tap is disabled! Use fully-scoped user/tap/formula naming instead.
$ brew install bazelbuild/tap/ibazel
==> Installing ibazel from bazelbuild/tap
<snip>

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,7 @@ If you run a mac you can install it from [homebrew](https://brew.sh).
 
 ```
 $ brew tap bazelbuild/tap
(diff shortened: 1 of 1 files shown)


achew22:
@BooneJS, thanks for sending the fix in and thanks for contributing!

PR #322, opened 2020-01-01
Ensure that all processes associated with group are terminated

mrmeku:
On Unix this involves finding the process group ID of the process
and sending a signal to kill the inverse of the that ID

On windows, this involves using the program TASKKILL and passing
the /T flag to indicate that the entire process tree should terminate.


achew22:
@mrmeku, wow, great PR! Out of curiosity does this mean that if you SIGKILL the iBazel task that the process group is now linked since you're setting the process group ID?

achew22:
@mrmeku anything I can do to help this across the finish line?

PR #340, opened 2020-02-05
support WORKSPACE.bazel file

alexeagle:
This was added in Bazel 2.0


achew22:
Thanks for the contribution!

joneshf:
What tends to be the turnaround time for releases? I'm not in a rush, but looking to know when this change might make it into the next release, so I can plan accordingly.

Issue #339, opened 2020-02-04
Try resolve bazel binary from @bazel/bazelisk installed locally


Toxicable:
Now that @bazel/bazelisk is deployed to npm https://www.npmjs.com/package/@bazel/bazelisk
We should try resolve the bazel binary from that local install.

achew22:
The file in question is [here](https://github.com/bazelbuild/bazel-watcher/blob/master/release/npm/index.js). The way I would go about developing it would be to `npm install @bazel/ibazel` and `@bazel/bazelisk` and then manually edit the `index.js` that is in the `ibazel` until it exhibits all the functionality you need. Then copy your changed `index.js` over into this repo and send a PR.

Is there anything else you need from me to get this done?

alexeagle:
Last time I made a related change, we did it in the go code:
https://github.com/bazelbuild/bazel-watcher/pull/275/files

I don't remember enough context to say which is better or more correct though

alan-agius4:
This is pretty similar to https://github.com/bazelbuild/bazel-watcher/issues/341 seems like the changes in https://github.com/bazelbuild/bazel-watcher/pull/275 have regressed.

PR #346, opened 2020-02-14
Support bazelisk

zoidbergwill:
Closes https://github.com/bazelbuild/bazel-watcher/issues/339

TODO:
- [x] Add tests

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -71,14 +71,18 @@ func bazelNpmPath(ibazelBinPath string) (string, error) {
 // /DIR/node_modules/@bazel/bazelisk/bazelizk-darwin_amd64
 func bazeliskNpmPath(ibazelBinPath string) (string, error) {
 	s := strings.Split(ibazelBinPath, "/")
-	for i := 0; i + 4 < len(s); i++ {
+	for i := 0; i+4 < len(s); i++ {
 		prefix, nm, scope, pkg, dir, bin := s[0:i], s[i], s[i+1], s[i+2], s[i+3], s[i+4]
 		if nm == "node_modules" && scope == "@bazel" && pkg == "ibazel" && dir == "bin" {
-			name := strings.Join(append(prefix, nm, scope, "bazelisk", "bazelisk-" + bin), "/")
+			var ext string
+			if strings.HasPrefix(bin, "windows_") {
+				ext = ".exe"
+			}
+			name := strings.Join(append(prefix, nm, scope, "bazelisk", "bazelisk-"+bin+ext), "/")
 			return name, nil
 		}
 	}
-	return "", errors.New("bazel binary not found in @bazel/bazelisk package")
+	return "", errors.New("bazelisk binary not found in @bazel/bazelisk package")
 }
 
 func findBazel() (string) {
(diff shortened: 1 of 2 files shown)


bazel/bazel.go
```diff
@@ -73,6 +91,11 @@ func findBazel() (string) {
 	if npmPath, err := bazelNpmPath(filepath.ToSlash(os.Args[0])); err == nil {
 		return filepath.FromSlash(npmPath)
 	}
+	// Frontend devs may have installed @bazel/bazelisk and @bazel/ibazel from npm
+	// If they also have bazelisk in the $PATH, we want to resolve this one, to avoid version skew
+	if npmPath, err := bazeliskNpmPath(filepath.ToSlash(os.Args[0])); err == nil {
```
achew22: You should probably check for Bazelisk before checking for Bazel.

bazel/bazel.go
```diff
@@ -73,6 +91,11 @@ func findBazel() (string) {
...
+		return filepath.FromSlash(npmPath)
+	}
 
 	// Check in $PATH for system-installed Bazel
 	if path, err := exec.LookPath("bazel"); err == nil {
```
achew22: Can you also check to see if Bazelisk is on the path?

zoidbergwill:
Thanks for the quick review @achew22 , those are great points. Updated.

bazel/bazel.go
```diff
@@ -63,17 +63,43 @@ func bazelNpmPath(ibazelBinPath string) (string, error) {
...
+func bazeliskNpmPath(ibazelBinPath string) (string, error) {
+	s := strings.Split(ibazelBinPath, "/")
+	for i := 0; i + 4 < len(s); i++ {
+		prefix, nm, scope, pkg, dir, bin := s[0:i], s[i], s[i+1], s[i+2], s[i+3], s[i+4]
+		if nm == "node_modules" && scope == "@bazel" && pkg == "ibazel" && dir == "bin" {
+			name := strings.Join(append(prefix, nm, scope, "bazelisk", "bazelisk-" + bin), "/")
+			return name, nil
+		}
+	}
+	return "", errors.New("bazel binary not found in @bazel/bazel package")
```
zoidbergwill: bazelisk, silly

```suggestion
	return "", errors.New("bazel binary not found in @bazel/bazelisk package")
```

bazel/bazel.go
```diff
@@ -63,17 +63,43 @@ func bazelNpmPath(ibazelBinPath string) (string, error) {
 	return "", errors.New("bazel binary not found in @bazel/bazel package")
 }
 
+// bazeliskNpmPath looks up a relative path to a binary from @bazel/bazelisk
+// This is used as an alternate resolution when no bazel binary is in the $PATH
+// When running from the @bazel/ibazel npm package, our binary is
+// /DIR/node_modules/@bazel/ibazel/bin/darwin_amd64/ibazel
+// We can find bazelisk in
+// /DIR/node_modules/@bazel/bazelisk/bazelizk-darwin_amd64
+func bazeliskNpmPath(ibazelBinPath string) (string, error) {
```
achew22: Could I have you add a test for this? It doesn't depend on any system calls or anything so a [table driven test](https://github.com/golang/go/wiki/TableDrivenTests) would be sufficient.

zoidbergwill: Sorry, added these earlier but couldn't figure out why I was getting a segfault, but fixed now.

Fortunately they caught me missing the `.exe` extension for Windows. I'm not 100% comfortable with how I'm doing the platform detection though.

PR #349, opened 2020-02-19
Enable run_output by default

achew22:
Since run_output_interactive is enabled, it should be safe to prompt the user
if they want to execute the program that was specified by their repo's
`.bazel_fix_commands.json`

### DIFF
--- a/ibazel/output_runner/output_runner.go
+++ b/ibazel/output_runner/output_runner.go
@@ -35,7 +35,7 @@ import (
 
 var runOutput = flag.Bool(
 	"run_output",
-	false,
+	true,
 	"Search for commands in Bazel output that match a regex and execute them, the default path of file should be in the workspace root .bazel_fix_commands.json")
 var runOutputInteractive = flag.Bool(
 	"run_output_interactive",


achew22:
@borkaehw + @alexeagle , could I get a sanity check on this? I believe this is safe to enable by default. In the event that you run in an untrusted repo, you would still have to pass `--run_output_interactive=false` in order to be vulnerable. That said, you're already running untrusted code on your computer so you're probably pwned already if you get to this safety check.

borkaehw:
I believe it's safe and nice to have.

achew22:
@alexeagle Is that the case for you as well?

achew22:
@alexeagle, any thoughts on this?

alexeagle:
Sorry I missed the notification - I don't really know how this feature works. "the default path of file should be in the workspace root" - why is there a file at all? Is a compiler tool supposed to create a file when it runs?

I'm certainly in favor of making it easier for compilers to get their fixes applied.

achew22:
@alexeagle interesting, have I misinterpreted the way that the quick fixes are implemented? I was under the impression they were scanning the stderr from the action and if they match some regex then they are executed. Is there a mechanism in a rule to have it emit something that says "The build is broken, please run this command?"

borkaehw:
@achew22 your understanding is correct. The regex is defined in a json file. The rule is responsible to emit the command, if the command matches the regex, run the command.

Issue #300, opened 2019-11-12
If the inner program crashes or dies, ibazel should restart it


achew22:
Right now if the inner program dies you get the error:

```
Error writing success to stdin: write |1: broken pipe
```

This means that the inner program closed its `stdin`, usually due to crashing/exiting, and needs to be restarted.

PR #366, opened 2020-03-22
Restart subprocess on changes if it crashed

mrmeku:
Fixes https://github.com/bazelbuild/bazel-watcher/issues/300

### DIFF
--- a/ibazel/command/notify_command.go
+++ b/ibazel/command/notify_command.go
@@ -45,7 +45,8 @@ func NotifyCommand(startupArgs []string, bazelArgs []string, target string, args
 }
 
 func (c *notifyCommand) Terminate() {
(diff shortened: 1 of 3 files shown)


ibazel/command/notify_command.go
```diff
@@ -89,6 +89,12 @@ func (c *notifyCommand) Start() (*bytes.Buffer, error) {
 }
 
 func (c *notifyCommand) NotifyOfChanges() *bytes.Buffer {
+	if !c.IsSubprocessRunning() {
```
achew22: I think we should also print out a log message informing the user that we are restarting their subprocess

mrmeku: Done

PR #374, opened 2020-04-19
Switch to cquery instead of query

### LINKED ISSUE #305: Proposal:  use bazel's cquery in place of query for queryForSourceFiles checks
Currently ibazel uses `bazel query` to determine if the build graph has changed.  However `bazel query` does not allow for configurations to be considered. 

If you set a `--define` flag for your `build`s, every time ibazel runs `bazel query`, the analysis cache is invalidated in determining the `query` results only to have the results invalidated again by the next `build`.

If `bazel cquery` is used instead, the `--define` flag is now considered/valid and the correct graph is created.  

`cquery` inherits all of its flags from `build` (as does `run` and `test`) which leads to all supported flags in ibazel being able to be forwarded to the "query" checks

---

Under the hood the commands that are called by ibazel in this instance look something like this:
```
Calls: bazel build //path/to/module:target
      ^ creates analysis cache using `--define` flag configuration

Change made to a file

Calls: bazel query //path/to/module:target
      ^ creates new analysis cache without `--define` flag, throwing out old cache
Calls: bazel build //path/to/module:target
      ^ creates new analysis cache using `--define` flag, throwing out old cache
```

with `cquery` it would instead be:
```
Calls: bazel build //path/to/module:target
      ^ creates analysis cache using `--define` flag configuration

Change made to a file

Calls: bazel cquery //path/to/module:target
      ^ uses analysis cache already created, updating as necessary
Calls: bazel build //path/to/module:target
      ^ uses analysis cache already created
```

achew22:
@aiuto, thanks so much for comaintanence on this project. It's really put a burst of wind into my sails! I can't tell you how excited I am to have you help.

Thanks for the review!

Fixes #305


aiuto:
You'll have to give me a day or two. This looks like something I am going to have to actually read carefully

achew22:
@aiuto friendly ping

aiuto:
I whacked on the settings yesterday.  I think you will be able to merge this.

achew22:
@aiuto thanks for the review!

PR #382, opened 2020-05-08
Allow `--build_tag_filters` and `--build_tests_only` to be specified 

devversion:
Proposal to allow the `--build_tag_filters` and `build_tests_only` Bazel flags.

Are there any guidelines on what flags should be in here or not? I consider these flags as commonly used, so it might make sense to have them inside here.


achew22:
Thanks so much for your contribution!

PR #384, opened 2020-05-11
Pass override repository flags to query commands and add e2e test

### LINKED ISSUE #383: ibazel with --override_repository can't handle changes in the build tree
If the overridden repository contains targets that the original repository doesn't, then it fails to build and watch correctly.

I believe this is because we don't pass the `--override_repository` flag to the bazel `query` and `cquery` commands.

lewish:
Fixes #383


achew22:
With an end to end test attached!? What a PR! Thanks so much for the fix

PR #394, opened 2020-05-24
Only post messages to STDIN if subprocess is running

mrmeku:


achew22:
Can you add an e2e test to validate the absence of the error output?

mrmeku:
> Can you add an e2e test to validate the absence of the error output?

Done! Good thing you asked too, I found a bug in `IsSubprocessRunning` as a result

e2e/example_client/BUILD.bazel
```diff
@@ -13,3 +14,12 @@ go_library(
         "@io_bazel_rules_go//go/tools/bazel_testing:go_default_library",
     ],
 )
+
+go_bazel_test(
```
achew22, line 18: The hope was that people would reuse `example_client` as a library and the other tests would live in `e2e/your_test_name` Could I ask you to `git mv` these up a dir? shouldn't be a painful move I hope

e2e/example_client/example_client_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"time"
+)
+
+func TestRestartProcess(t *testing.T) {
+	client := StartLiveReload(t)
+	defer client.Cleanup()
+	client.Kill(t)
+
+	// Wait for the server to kill itself (it has a 1 second delay)
+	time.Sleep(time.Second * 2)
```
achew22, line 15: Would it be better to in `client.Kill()` try to get the `/config` url over and over again until a connection error? Then there wouldn't need to be a sleep in here

mrmeku: Done

mrmeku:
@achew22 The command unit tests were relying upon the bug in `IsSubprocessRunning` to pass.

I'm not sure how to resolve this without explicitly mocking `process_group.ProcessGroup`

Issue #333, opened 2020-01-27
Stop running subprocesses as soon as changes are detected


borkaehw:
Not sure if it's a bug or a design in mind.

We use iBazel to run a tcp server. When we make a change, iBazel behaves like
1. detect changes
2. build new code
3. stop old server
4. start new server

This is not ideal for us. Building new code could take few minutes. Developers can still hit the old server, which could be confusing sometimes to see things are not changing. The only way to tell is by looking at the console. It would be ideal if iBazel can stop the server right away the changes have been detected.

We tried to identify how iBazel starts and stops subprocesses but no luck. we really help some helps. Thanks.

achew22:
There is a fine line between bug and feature. I use this feature regularly in my testing where I want to see the results of the current test finish, then as soon as the test finishes restart the compile/test loop.

That said, I don't think there is any reason we can't tag binaries with something like `IBAZEL_SHORT_CIRCUIT_RESTART` or something like that. Would you be interested in having that functionality?

borkaehw:
That would be great. It sounds like what we are expecting. How complex would it be to implement it?

achew22:
If you wanted to modify the code here to introduce a new switch for short circuiting, that would be a good place for it. I don't think it would be extremely difficult, the hardest part would probably be writing an e2e test for the case where that tag isn't applied.

https://github.com/bazelbuild/bazel-watcher/blob/e695e93213e803ca4091f39ef121fbddb2d39c42/ibazel/ibazel.go#L389

mrmeku:
@achew22 I think that this might be a sane default for binaries tagged with `ibazel_live_reload`

What do you think?

achew22:
I'm not sure this is a good default for `ibazel_live_reload`. The major use for live reload is when you're trying to rapidly iterate on your frontend. If the frontend is going to reload every time then you've got to persist the dev server state because the server will lose it every time it comes up. It also means if you're doing a Java server or something with long start up times then you're going to have to wait 30s on every file change to see results.

Maybe I'm missing something, can you help me understand the idea better?

mrmeku:
Assume that a typical javascript rebundling cycle takes ~8 seconds

A common pattern I find myself in is fixing a small error, saving, realizing a made a small syntax error, re-saving and then having to wait 16 seconds to see if my code compiles

[...]

PR #396, opened 2020-05-25
Add support for ibazel_short_circuit tag

mrmeku:
Fixes https://github.com/bazelbuild/bazel-watcher/issues/333


ibazel/ibazel.go
```diff
@@ -339,26 +353,31 @@ func verb(s string) string {
 }
 
 func (i *IBazel) build(targets ...string) (*bytes.Buffer, error) {
-	b := i.newBazel()
+	if i.bazel != nil {
```
achew22, line 356: This kills the `i.bazel` indiscriminately, should there be a check to make sure short circuiting is enabled?

process_group/process_group_windows.go
```diff
@@ -41,6 +41,10 @@ func (pg *winProcessGroup) RootProcess() *exec.Cmd {
 	return pg.root
 }
 
+func (pg *winProcessGroup) Run() *exec.Cmd {
```
achew22, line 44: The type in `process_group_unix.go` and in `process_group_windows.go` do not match,

```
func (pg *unixProcessGroup) Run() error {
```

vs

```
func (pg *winProcessGroup) Run() *exec.Cmd {
```

ibazel/ibazel.go
mrmeku: done, thanks for catching that!

process_group/process_group_windows.go
mrmeku: done

PR #399, opened 2020-05-30
Add log messages to explain iBazel time usage

karsmars:
Sometimes IBazel takes a long time when running `bazel info`. This commit simply adds a couple console logs so developers know what is going on and what is taking time.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -232,17 +232,21 @@ func (b *bazel) Info() (map[string]string, error) {
 	b.WriteToStdout(false)
 	stdoutBuffer, _ := b.newCommand("info")
 
-	// This go function only prints if 'bazel info' takes longer than 5 seconds
-	runFinished := false
+	// This gofunction only prints if 'bazel info' takes longer than 8 seconds
+	doneCh := make(chan struct{}, 1)
+	defer func() {
+		doneCh <- struct{}{}
+	}()
 	go func() {
-		time.Sleep(5*time.Second)
-		if !runFinished{
-			log.Logf("Running 'bazel info'...")
+		select {
+			case <- doneCh:
+				// Do nothing since we're done.
+			case <- time.After(8*time.Second):
+				log.Logf("Running `bazel info`... it's being a little slow")
 		}
 	}()
-	
+
 	err := b.cmd.Run()
-	runFinished = true
 	if err != nil {
 		return nil, err
 	}
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -30,6 +30,7 @@ import (
 	"github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf/blaze_query"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/bazelbuild/bazel-watcher/ibazel/log"
 )
 
 var bazelPathFlag = flag.String("bazel_path", "", "Path to the bazel binary to use for actions")
@@ -229,7 +230,7 @@ func (b *bazel) Info() (map[string]string, error) {
 	b.WriteToStderr(false)
 	b.WriteToStdout(false)
 	stdoutBuffer, _ := b.newCommand("info")
-
+	log.Logf("Running 'bazel info'...")
 	err := b.cmd.Run()
 	if err != nil {
 		return nil, err
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -25,6 +25,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf/analysis"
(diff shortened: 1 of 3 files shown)


bazel/bazel.go
```diff
@@ -229,7 +230,7 @@ func (b *bazel) Info() (map[string]string, error) {
 	b.WriteToStderr(false)
 	b.WriteToStdout(false)
 	stdoutBuffer, _ := b.newCommand("info")
-
+	log.Logf("Running 'bazel info'...")
```
achew22, line 233: What would you think about starting a goroutine and only printing this out if it takes a non-trivial amount of time to execute?

karsmars: Good idea. I'll add that in.

karsmars:
I'm in the process of testing my changes, I will re-request review when they are ready.

bazel/bazel.go
karsmars: Let me know what you think of the changes or how you would implement it :)

bazel/bazel.go
```diff
@@ -230,7 +232,17 @@ func (b *bazel) Info() (map[string]string, error) {
 	b.WriteToStdout(false)
 	stdoutBuffer, _ := b.newCommand("info")
 
+	// This go function only prints if 'bazel info' takes longer than 5 seconds
+	runFinished := false
```
achew22, line 236: Thank you so much for all the hard work you clearly put into this! I'm delighted that you're so willing to jump into the code and get dirty. That's real open source right there!

Go has some really great synchronization primitives for exactly this usecase which I think will make it a little more clear what is going on. If you'll afford me a moment to talk about them, I'll attempt to describe them as best I can in comments. Feel free to copy/paste this code into the repo as a part of your commit.

Go has a concept called `defer` where it will execute a function of your choosing as soon as the function you're inside of terminates. So, for example, you can run:

```
defer func() {
  fmt.Printf("Hello")
}()
fmt.Printf(" world")
```

which will have the very counter intuitive output of ` worldHello`. Try it out [here](https://play.golang.org/p/LriR75Il-nk).

Additionally, Go has a concept of channels, which are first in, first out (FIFO) pipes. You can define a channel like this:

```
ch := make(chan bool, 1)
```

That is a `chan`nel that holds `1` element in it before blocking. You can write into that by using the syntax `ch <- true`, and assuming the channel isn't already full (remember it can only hold 1 element at most) it will immediately continue on.

The consumer of the `chan`nel can read the value (or read and ignore the value). To read the value you would run `myBoolVar := <- ch`. To discard the value, you would run just `<- ch`. If there is no value to be read the goroutine that is reading from the channel will block until a value is added into the `chan`nel.

The time package in go also has a cool method, [`time.After`](https://golang.org/pkg/time/#After) that returns a `chan`nel of `time.Time` after the provided time has elapsed. This means you can use that channel to block until a task is complete. Sorry I couldn't think of a descriptive codesample that would demo this.

The final go concept is called a `select` which will try to pull a value out of multiple channels and do the first one that responds, blocking if it can't read from any.

```
select {
case <-time.After(1 * time.Second):
	fmt.Printf("1")
case <-time.After(2 * time.Second):
	fmt.Printf("2")
case <-time.After(3 * time.Second):
	fmt.Printf("3")
}
```

You'll see that you get the output "1", because the channel created by `time.After` is going to be the first to fire. You can see what this looks like [here](https://play.golang.org/p/8WQXI1pR73I).

Now when you take all of those things combined with the `go`routine that you already figured out how to do, you can make a "done" channel, kick off a go routine that will watch it (and a time.After)

You can see what this looks like [here](https://play.golang.org/p/mc5J2VJEqqd). Try messing with the sleep statement below and see what happens.

putting all of this together you should get something like this:

```
doneCh := make(chan struct{}, 1)
defer func() {
  doneCh <- struct{}{}
}()
go func() {
  select {
    case <- doneCh:
      // Do nothing since we're done.
    case <- time.After(5*time.Second):
      log.Logf("Running `bazel info`... it's being a little slow")
  }
}()
```

If you ever get the hankering to do some go programming, the repo is here and I'm always happy to describe how I would approach a problem like this. Or, if you think I'm boring and tedious, feel free to ignore me 😄 

Hope this helps. Cheers!

karsmars: Thank you for this above-and-beyond response! It's clear you took some of your personal time just to help me learn. I'll implement some of the suggestions you made.

karsmars: I had learned a bit about some of the structures you described above, but your explanation and examples helped me understand them a lot better. And I have no improvements upon your suggested code. Thank you so much!

achew22:
@kalbasit thanks so much for your contribution. It may seem small, but usability improvements like this add up to become wonderful things. I also appreciate you affording me the chance to wax poetic about my implementation ideas. I look forward to working with you more in the future!

kalbasit:
> @kalbasit thanks so much for your contribution. It may seem small, but usability improvements like this add up to become wonderful things. I also appreciate you affording me the chance to wax poetic about my implementation ideas. I look forward to working with you more in the future!

Did you mean @karsma?

PR #404, opened 2020-06-16
Add --dynamic_mode flag to overridable list

akonradi:
Per the manual [1], Bazel supports disabling dynamic linking via the --dynamic_mode flag. Allow overriding that for iblaze builds.

[1] https://docs.bazel.build/versions/master/user-manual.html#flag--dynamic_mode


achew22:
Thanks for your contribution! This will go out in the next release

Issue #245, opened 2019-04-05
ibazel run on container_image does not work


Globegitter:
I have a container_image like follows:

```
container_image(
    name = "image",
    base = "@nginx_alpine//image",
    directory = "/etc/nginx/",
    docker_run_flags = "--rm -it -p 8080:8080",
    files = [
        "nginx.conf",
    ],
    legacy_run_behavior = False,
)
```

and all works well when doing `bazel run :image`, it starts the image, i can CTRL+C out of it etc. but when using `ibazel run` I am getting the error message from docker:

```
the input device is not a TTY
```

By removing either the `-i` or `-t` in the `docker_run_flags` I can get around the error message but then pressing CTRL+C does not actually stop the docker container and on rebuild then it complains that the port is already allocated.

Not sure if there is an easy way to fix this and also happy to investigate myself (though will not happen in the next weeks) but wanted to make sure it is tracked.

statik:
I just ran into this as well - would love to migrate my project off of docker compose and ibazel run of container targets would be just the ticket.

achew22:
Sounds like a great opportunity for a first PR. How can I help you with that?

statik:
I'm happy to try and narrow down the issue, I think the first step would be for me to make a little standalone repo that illustrates the problem, and then it will take some investigating to try and narrow down why the process is not getting killed.

Globegitter:
@statik My experience with this sort of thing has been that there is an issue with correct signal propagation. I have ran into a similar issue in the past with rules_nodejs where they wrapped all binaries in a shell script. So it could be a similar issue here. Not sure about the TTY error message though, I do not remember seeing that in the case of rules_nodejs.

statik:
I made a test case here: https://github.com/kindlyops/bazel-cross-docker

Need to do some checking to see what signals are making it through.

achew22:
@statik did you have any luck diagnosing which signals were getting eaten?

statik:
@achew22 no, I haven't been able to make any progress with debugging unfortunately.

vanils:
@statik @achew22 The issue seems to be ibazel sending `SIGKILL` to terminate a process on Unix. In case of Docker this doesn't work well. `SIGKILL` simply kills the Docker CLI giving no chance to proxy any signals to the container which is therefore left up and running. As a result there is _the port already allocated_ error when attempting to start a server without stopping the previous instance of it.

[...]

PR #405, opened 2020-06-28
Using SIGTERM to stop the subprocess

vanils:
Currently ibazel uses `SIGKILL` to kill the subprocess. This doesn't work properly when the subprocess must be gracefully shutdown - like the case with Docker (#245). Changing this signal to `SIGTERM` will fix the issue. There is still `SIGKILL` used as a fallback if `SIGTERM` doesn't kill the subprocess in a reasonable time.


vanils:
Seems like I created a race condition on unit tests with the latest changes. Weird that locally tests passed. Will fix that and then remove a draft flag

ibazel/command/notify_command.go
```diff
@@ -50,17 +52,18 @@ func (c *notifyCommand) Terminate() {
 		return
 	}
 
-	// Kill it with fire by sending SIGKILL to the process PID which should
-	// propagate down to any subprocesses in the PGID (Process Group ID). To
-	// send to the PGID, send the signal to the negative of the process PID.
-	// Normally I would do this by calling c.cmd.Process.Signal, but that
-	// only goes to the PID not the PGID.
-	c.pg.Kill()
+	c.pg.Kill(c.signum)
```
achew22, line 55: 1) I think we should probably rename the `Kill` command to something like `Signal` since you could send something like a `SIGHUP`, or `SIGUSR1` which are not kill commands.

2) Out of curiosity, what if we did something here this:

```
c.pg.Signal(SIGTERM)
var done chan bool
go func() {
  c.pg.Wait()
  done <- true
}
select {
case <- time.After(1*time.Second):
  c.pg.Signal(SIGKILL)
case <- done:
  // No action necessary since the job is gone on its own.
}
c.pg.Close()
```

I think that would handle the case where the process needs a small amount of time to shut down and also kill it if it really needs to die without adding a flag.

Can you tell me why my idea isn't sufficient, I think I might be missing something but I'm not sure?

ibazel/ibazel.go
```diff
@@ -41,6 +41,11 @@ var osExit = os.Exit
 var bazelNew = bazel.New
 var commandDefaultCommand = command.DefaultCommand
 var commandNotifyCommand = command.NotifyCommand
+var exitMessages = map[os.Signal]string{
```
achew22, line 44: Much cleaner than what I did. Good job!

ibazel/ibazel.go
```diff
@@ -124,43 +132,47 @@ func New() (*IBazel, error) {
 func (i *IBazel) handleSignals() {
 	// Got an OS signal (SIGINT, SIGTERM, SIGHUP).
 	sig := <-i.sigs
+	i.isClosing = true
```
achew22, line 135: I don't think this is quite correct. `isClosing` should only be set to true when the subprocess is no longer running and ibazel receives a SIG*. Also, I think I would prefer using the preexisting state of [QUIT](https://github.com/bazelbuild/bazel-watcher/blob/7bbfce42b1667710a3bd266e82cbfd22c3b88979/ibazel/ibazel.go#L54) to track this. If you set the state to QUIT the next call to `iteration` won't have any action to perform which is the same as putting in a boolean like this.

vanils: I completely missed this QUIT state. I'll play around with this. The reason why this .isClosing state is set here is to make sure no more loop actions are taken after this point. Let's say termination of the subprocess takes longer, like 5 seconds. That's enough for a user to still make a change on a file and trigger restart of a subprocess. But now as I wrote this, restart means it needs to terminate first, which would mean QUIT state gets assigned before starting a new process. I think you are absolutely right, I'll try it out.

ibazel/command/notify_command.go
vanils: 1. Makes sense. I'll rename. `Signal` sounds good to me.

2. I have played around with this idea as well. Technically it shouldn't cause any issues. Not even with a case of a Docker image, as Docker already got time to proxy the first signal to a container.

The downside of it would be to remove a control from a user. As an example, I'm personally planning to setup a development environment as close to a production environment as possible. That means using a Docker containers and graceful shutdown of servers. For that reason checking that graceful shutdown works as expected is important. If I would see operation stopping before it's completed, I would have to start investigating it. After a while it would turn out to be ibazel and I would be disappointed not being able to do anything about it.

This is a reason why I didn't implement it this way when ibazel detects SIGINT. SIGTERM and SIGHUP would actually result in a behavior like this. Instead of 1 second I used 10 seconds though. Now that I saw you using 1 second here in an example I started to think if 10 seconds was a bit too much?

achew22: @vanils, thanks for the thoughtful response.

> The downside of it would be to remove a control from a user. As an example, I'm personally planning to setup a development environment as close to a production environment as possible. That means using a Docker containers and graceful shutdown of servers. For that reason checking that graceful shutdown works as expected is important. If I would see operation stopping before it's completed, I would have to start investigating it. After a while it would turn out to be ibazel and I would be disappointed not being able to do anything about it.

Hrm. These are very reasonable concerns. The only reason I'm pushing back on this is because right now the only control surfaces we have are basically to enable testing ibazel or to help develop ibazel in some other way. I don't believe you can even run -h to get a list of most of these flags.

* `--debounce` - Only intended to be used in tests to make them run faster
* `--log_to_file` - Used in testing to validate that the log output matches expectations
* `--nolive_reload` - Used to simplify some of the non-live-reload testcases (primarily controlled through bazel tags)
* `--bazel_path` -- Used in e2e testing to override the bazel path for when I was testing against many versions of bazel
* `--profile_dev` - Used to profile ibazel to find bottlenecks

There is really only one flag that does modifies control flow, `--run_output`. It exists because we can't defer trust to the repo. For example you clone a potentially malicious repo and you wanna see what happens when you do X, `ibazel` shouldn't prohibit you from doing things safely if you'd like to opt into it.

> This is a reason why I didn't implement it this way when ibazel detects SIGINT. SIGTERM and SIGHUP would actually result in a behavior like this. Instead of 1 second I used 10 seconds though. Now that I saw you using 1 second here in an example I started to think if 10 seconds was a bit too much?

What if we printed out an error message that said that the process wasn't killed within the X second window? Would that provide you the shutdown latency visibility you're seeking?

vanils: Ahh. So maybe I missed the point of the your comment. So there would be 2 changes suggested.

- Use SIGTERM by default. Maybe removing the need for `--signal` flag in the first place?
- Use this timeout behavior with SIGINT, not just when detecting SIGTERM and SIGHUP?

Did I got it right? Your reasoning sounds good to me.

> What if we printed out an error message that said that the process wasn't killed within the X second window? Would that provide you the shutdown latency visibility you're seeking?

That sounds good. Maybe we could use higher timeout than 1 second? A user could always bail out faster by repeatedly hitting ctrl-c if they want. Maybe something like 5 seconds or even 10 seconds?

achew22: > * Use SIGTERM by default. Maybe removing the need for --signal flag in the first place?

Anything to avoid flags 😉 

> * Use this timeout behavior with SIGINT, not just when detecting SIGTERM and SIGHUP?
> Did I got it right? Your reasoning sounds good to me.

Yep, I think that provides the needed task killing and also keeps the ergonomics of the current set up. Doing the right thing effectively silently (TM) the vast majority of the time is the goal here.

> That sounds good. Maybe we could use higher timeout than 1 second? A user could always bail out faster by repeatedly hitting ctrl-c if they want. Maybe something like 5 seconds or even 10 seconds?

Good point. The only issue with 10s is that it will mean we need to wait 10s in an e2e test to verify it and 10s is a long time for a test...

And now to completely contradict what I said just a few paragraphs ago... since that's controlling behavior that should be manipulated only in tests, would you be okay turning that into a flag but having a sane default of 10s?

vanils: > And now to completely contradict what I said just a few paragraphs ago... since that's controlling behavior that should be manipulated only in tests, would you be okay turning that into a flag but having a sane default of 10s?

Sounds good! :) I'll mark this pr as a draft again and undraft it when all the changes have been done.

ibazel/ibazel.go
vanils: > isClosing should only be set to true when the subprocess is no longer running and ibazel receives a SIG*.

Hey still one more question. With this PR I would suggest changing a behavior of ibazel to automatically exit when the subprocess exits. The reasoning can be found from the original PR description. In a nutshell: it would be one less ctrl-c for user and maybe a bit simpler implementation. Does that sound good? Just double checking as now when rereading this part it would suggest maybe you missed this? I can also bring back this double-SIGINT-to-exit behavior if it's preferred?

achew22: > Hey still one more question. With this PR I would suggest changing a behavior of ibazel to automatically exit when the subprocess exits. The reasoning can be found from the original PR description. In a nutshell: it would be one less ctrl-c for user and maybe a bit simpler implementation. Does that sound good? Just double checking as now when rereading this part it would suggest maybe you missed this? I can also bring back this double-SIGINT-to-exit behavior if it's preferred?

To me, the double-SIGINT-to-exit is the preferred default behavior. I think this would mean that `ibazel test` quits after every invocation finishes which is definitely not the correct behavior. If you would like to change that behavior I think we can do that in a separate PR and make it a tag that's read off the target you're running.

WDYT?

vanils: Thanks for a quick reply :) I wasn't accurate with my comment. Ibazel would **only** exit if the subprocess was terminated due to SIG* on ibazel. Therefore `ibazel test` keeps running until the first ctrl-c (or SIGTERM/SIGHUP).

Does it sound better now? Usually a change like this would definitely be a separate pr. The only reason why I embedded it here was that it was in my opinion the simplest and cleanest solution so solve:

> it would remove the need to figure out which signal to use for example in scenario where user hits ctrl-c, then makes a change in source files which would restart a subprocess, then hits ctrl-c again

This case can be figured out the other way though. One that I considered was resetting `.interruptCount` on every restart (currently this is never reseted). That would force the flow of a subprocess termination to always be SIGTERM - SIGKILL - exit ibazel due to three SIGINT. This sounds a bit risky though. What if user causes loop which would keep modifying source files which would then again cause ibazel to restart again and again. Now bailing out ibazel would be a matter if you are fast enough to hit that SIGINT three times before ibazel gets restarted.

The suggested solution sounded better to me, maybe I'm still missing something essential? 🤔 Nevertheless, I don't have a strong opinion on this. If you still prefer double-SIGINT-to-exit then I will make it work here.

achew22: Ah, yes. The single-SIGINT-to-exit-the-inner-process is something that I do use on a regular basis. I use it to restart devservers when I've changed the server side code and they are set to `live_reload`. Would it be okay if we open another issue and discuss the merits of that change separately?

Sorry to be a PITA

vanils: No problem. :) I actually kind of assumed this change of double-SIGINT won't pass as there must be things I haven't taken into account. Maybe I open another PR about this at some point, but let's follow now the same flow as it used to be. Thanks for the quick response again. 👍

vanils:
@achew22 I think this is ready for an another review. Thank you very much for the first round by the way :)

e2e/termination/termination_test.go
```diff
@@ -0,0 +1,93 @@
...
+	})
+}
+
+func TestTerminationBasic(t *testing.T) {
+	ibazel := e2e.SetUp(t)
+	e2e.MustWriteFile(t, "termination.sh", "printf \"Started 1!\";"+signalHandler)
+	ibazel.Run([]string{}, "//:termination")
+	ibazel.ExpectOutput("Started 1!")
+	e2e.MustWriteFile(t, "termination.sh", "printf \"Started 2!\";"+signalHandler)
+	if runtime.GOOS == "windows" {
```
achew22, line 72: Could you add a comment describing why you need to branch the validation logic?

ibazel/command/command.go
```diff
@@ -86,3 +97,27 @@ func subprocessRunning(cmd *exec.Cmd) bool {
...
+func terminate(pg process_group.ProcessGroup) {
+	pg.Signal(syscall.SIGTERM)
+	done := make(chan bool, 1)
+	go func() {
+		select {
+		case <-time.After(*gracefulDuration):
+			log.Logf("The subprocess wasn't terminated within %s", *gracefulDuration)
+			sendKillSignal(pg)
+		case <-done:
+			// The subprocess got terminated with SIGTERM
```
achew22, line 110: nit: s/got/was/

ibazel/command/command.go
```diff
@@ -86,3 +97,27 @@ func subprocessRunning(cmd *exec.Cmd) bool {
...
 	return true
 }
+
+func terminate(pg process_group.ProcessGroup) {
+	pg.Signal(syscall.SIGTERM)
+	done := make(chan bool, 1)
+	go func() {
+		select {
+		case <-time.After(*gracefulDuration):
+			log.Logf("The subprocess wasn't terminated within %s", *gracefulDuration)
```
achew22, line 107: Can you add information about what is going to happen now that it wasn't terminated within %s?

Something like "The subprocess wasn't terminated within %s and was killed forcefully?" Of course that would require putting it on the line after `sendKillSignal`

ibazel/command/default_command.go
```diff
@@ -43,19 +46,26 @@ func DefaultCommand(startupArgs []string, bazelArgs []string, target string, arg
...
-
-	// Kill it with fire by sending SIGKILL to the process PID which should
-	// propagate down to any subprocesses in the PGID (Process Group ID). To
-	// send to the PGID, send the signal to the negative of the process PID.
-	// Normally I would do this by calling c.cmd.Process.Signal, but that
-	// only goes to the PID not the PGID.
-	c.pg.Kill()
-	c.pg.Wait()
-	c.pg.Close()
+	if c.terminating {
```
achew22, line 53: Have you considered using a [`sync.Once`](https://golang.org/pkg/sync/#Once) to perform this work instead of tracking the state in a bool? It is designed for exactly this purpose.

ibazel/command/notify_command.go
```diff
@@ -45,20 +47,26 @@ func NotifyCommand(startupArgs []string, bazelArgs []string, target string, args
 }
 
 func (c *notifyCommand) Terminate() {
-	if c.pg == nil || !subprocessRunning(c.pg.RootProcess()) {
+	if !c.IsSubprocessRunning() {
```
achew22, line 50: I think a [`sync.Once`](https://golang.org/pkg/sync/#Once) would be an effective strategy here as well

ibazel/ibazel.go
```diff
@@ -324,6 +325,7 @@ func (i *IBazel) iteration(command string, commandToRun runnableCommand, targets
 		log.Logf("%s %s", strings.Title(verb(command)), joinedTargets)
 		i.beforeCommand(targets, command)
 		outputBuffer, err := commandToRun(targets...)
+		i.interruptCount = 0
```
achew22, line 328: Heh, this is subtle. I bet you pulled your hair out trying to figure out why it killed after just one `ctrl-c` when you hit this the first time 😉 sorry!

vanils: Luckily I'm bald 😄

ibazel/command/command.go
vanils: Currently logs look like this:

```
iBazel [4:43PM]: The subprocess wasn't terminated within 10s
iBazel [4:43PM]: Sending SIGKILL to the subprocess

iBazel [4:43PM]: Subprocess killed from getting SIGINT (trigger SIGINT again to stop ibazel)
```

To me it feels important having logs in chronological order. First comes a timeout, then a signal and then at some point the subprocess goes down.

How about "The subprocess wasn't terminated within %s. Forcing to close"?

Then it would be:

```
iBazel [4:58PM]: The subprocess wasn't terminated within 10s. Forcing to close.
iBazel [4:58PM]: Sending SIGKILL to the subprocess

iBazel [4:58PM]: Subprocess killed from getting SIGINT (trigger SIGINT again to stop ibazel)
```

What do you think?

ibazel/command/default_command.go
vanils: `sync.Once` seems like better approach. 👍

vanils:
@achew22 Fixed issues mentioned. Thank you for comments! :) One question about "The subprocess wasn't terminated within %s and was killed forcefully" above, but other than that I just made changes as you suggested.

ibazel/command/command.go
achew22: You're right. When considering the log output holistically your choice is superior.

ibazel/command/command.go
```diff
@@ -16,25 +16,36 @@ package command
...
 	"github.com/bazelbuild/bazel-watcher/ibazel/process_group"
 )
 
-var execCommand = process_group.Command
-var bazelNew = bazel.New
+var (
+	execCommand      = process_group.Command
+	bazelNew         = bazel.New
+	gracefulDuration = flag.Duration(
+		"graceful_duration",
```
achew22, line 38: Nit: since this is a flag people will never have to type, only tests I'm inclined to make it painfully obvious what it is doing. WDYT about renaming it to `gracefull_termination_wait_duration`?

ibazel/command/default_command.go
```diff
@@ -43,21 +45,22 @@ func DefaultCommand(startupArgs []string, bazelArgs []string, target string, arg
...
-	c.pg.Kill()
-	c.pg.Wait()
-	c.pg.Close()
+	c.termSync.Do(func() {
+		terminate(c.pg)
+	})
 	c.pg = nil
 }
 
+func (c *defaultCommand) SendKillSignal() {
```
achew22, line 58: Nit: `s/SendKillSignal/Kill`. This will read as `command.Kill()`which I believe is more descriptive because in windows it is not sending a kill signal.

ibazel/command/notify_command.go
```diff
@@ -45,22 +46,22 @@ func NotifyCommand(startupArgs []string, bazelArgs []string, target string, args
...
-	c.pg.Kill()
-	c.pg.Wait()
-	c.pg.Close()
+	c.termSync.Do(func() {
+		terminate(c.pg)
+	})
 	c.pg = nil
 }
 
+func (c *notifyCommand) SendKillSignal() {
```
achew22, line 59: Same naming here, `Kill` would be preferred IMO

achew22:
Thanks so much for all your hard work! Just a couple of global find and replaces. Once this is merged, I'd be very happy to merge and do a release so the whole world can enjoy your enhancements.

ibazel/command/command.go
vanils: This is the inner struggle of choosing between what's right and what feels good. While shorter names feel good the more descriptive names are the right thing to do. I agree!

I was thinking to make departure between variable name and flag name as the variable name does have a context so maybe it doesn't have to be that long. Maybe just `waitDuration` would be sufficient.

I don't speak english as my native language but isn't it "graceful" with one "l"? At least that's what Google tells me. :) Changing name to `graceful_termination_wait_duration` with a cautious assumption that I would be right about the spelling.

vanils:
Happy to contribute and thank you very much for reviews!

By the way: it seems so that //e2e/output_runner:go_default_test test have some random failures. The latest commit caused this test to fail on windows and I was unable to reproduce this locally no matter what I did. I ended up with the conclusion the failure wasn't caused by the latest commit and rerun tests (therefore the forcepush where I altered a commit message a bit). Now tests are green.

Failures can be found from https://buildkite.com/bazel/bazel-watcher/builds/1317

vanils:
@achew22 Any news about this? It’s been almost 2 months and still not merged 😕

Issue #402, opened 2020-06-08
MacOS: Error watching file "..." error: too many open files in system


joeljeske:
Same as https://github.com/bazelbuild/bazel-watcher/issues/109:

On macOS, we have an issue where a package and its deps could lead to too many open files. My understanding is that the FD limit is already raised to the max in-process by ibazel, so thereis nothing else I can raise to ignore this issue for now. 

Potential Solutions: 

* Flag to ignore packages: A flag could be introduced that would essentially ignore packages or package trees from the watch list. This would likely be the easiest but also rather annoying to work with. 
* Use FSEvents on MacOS: It is my understanding that the [FSNotify impl](https://github.com/fsnotify/fsnotify) that this project uses does not yet support FSEvents on MacOS, rather it uses kqueues. This results in the need to open a FD for each file to be watched. It is rather unperformant to do this for large amounts of files, and results in tripping the FD system limit. A better implementation is to use FSEvents on MacOS instead. That project does have a ["planned" item to use FSEvents](https://github.com/fsnotify/fsnotify/issues/11) however, it is nearing its 6th birthday and without much progress. It seems like the right approach is to add support for FSEvents in the FSNotify project, but it does not appear to be making any progress. 
* Refactor Code Base: It could be argued that my codebase is organized in such a way that it has too many dependencies.  This does not feel like the correct answer, as it puts the burden on every project to break dependencies to bypass this restriction. Also, it seems that the bazel mentality of encouraging building from source results is a deeper dependency tree that speeds up arriving at a maximum file-watch count.

achew22:
@joeljeske, thanks for leaving the well thought through issue. Response as inline as github will allow me. 

> On macOS, we have an issue where a package and its deps could lead to too many open files. My understanding is that the FD limit is already raised to the max in-process by ibazel, so thereis nothing else I can raise to ignore this issue for now.

Yes, this appears to be a hard limit. ibazel is setting it as high as it can but that doesn't seem to be high enough based on your usage.

> Flag to ignore packages: A flag could be introduced that would essentially ignore packages or package trees from the watch list. This would likely be the easiest but also rather annoying to work with.

Agreed that this is both simple, but also not particularly great from an ergonomics standpoint. I'd wager that there are a significant number of issues opened where people say "it isn't watching this directory" and I'm left scratching my head because they forget to mention that they set a flag somewhere that ignores dirs.

[...]

PR #406, opened 2020-06-29
integrate fsevents for macos

joeljeske:


achew22:
Hey Joel, unfortunately I'm not able to look at code at all until there is a CLA signed. I'm really excited to work through this problem with you so please ping the bug once you've settled things with the CLA bot.

Thanks!

joeljeske:
@googlebot I fixed it.

joeljeske:
@achew22 sorry about that, I didn't realize it was required until merge-time. It is fixed now. Thanks

joeljeske:
@achew22 I left a lot of notes and potential changes in #402 if you wouldn't mind looking at that

achew22:
Looking through this PR quickly it looks very reasonable. Sorry to have left tech debt for you to clean up! Whenever you want a full review tag it as ready and I'll take a real pass

ibazel/source_event_handler.go
```diff
@@ -1,47 +0,0 @@
...
-
-func (s *SourceEventHandler) Listen() {
-	for {
-		select {
-		case event := <-s.SourceFileWatcher.Events():
-			s.SourceFileEvents <- event
-
-			switch event.Op {
-			case fs_watcher.Remove, fs_watcher.Rename:
-				s.SourceFileWatcher.Add(event.Name)
```
joeljeske, line 34: @achew22 This file was deleted. As far as I could tell, this was created back when Ibazel watched individual files directly instead of their parent dirs. Now that it watches the parent dirs, I think the concern of IDEs deleting and replacing files is no longer valid. In my testing, it appeared to work correctly.

ibazel/fs_watcher_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"os"
+
+	"github.com/bazelbuild/bazel-watcher/ibazel/fs_watcher"
+	"github.com/bazelbuild/bazel-watcher/ibazel/fsnotify_watcher"
+	"github.com/bazelbuild/bazel-watcher/ibazel/fsevents_watcher"
+)
+
+func NewWatcher() (fs_watcher.Watcher, error) {
+	flag, ok := os.LookupEnv("IBAZEL_USE_LEGACY_WATCHER")
```
joeljeske, line 28: @achew22 I'm open to alternatives here if think it it would be better to default to the legacy, but as far as I can tell, its a positive to always use FSEvents over kqueue on macOS.

ibazel/BUILD
```diff
@@ -20,7 +20,11 @@ go_binary(
     importpath = "github.com/bazelbuild/bazel-watcher/ibazel",
     # Since we don't have any cgo dependencies, force build to be pure
     # golang. This will make multi-os compatibility *MUCH* easier.
-    pure = "on",
+    pure = "auto",
```
joeljeske, line 23: @achew22 I'm no go expert, so I do not really no the ramifications of toggling this, but I needed to do so in order to 
pull in FSEvents since it has C components

joeljeske:
@achew22 I'm having some trouble greening up the tests. I'm not getting valuable information back, if you have any insight, that would be great. I debug more on the tests, it might be related to flipping cgo for FSEvents? 

Anyway, the bulk of it should be ready for review if you wouldn't mind taking a look.

release/github/publish.sh
```diff
@@ -59,9 +66,9 @@ compile() {
 }
 
 # Now compiler ibazel for every platform/arch that is supported.
-compile "linux"   "amd64"
-compile "darwin"  "amd64"
-compile "windows" "amd64"
+compile "linux"   "amd64"  ""
+compile "darwin"  "amd64"  "cgo"
```
joeljeske, line 70: @achew22, I do not know exactly where this is built and am not super familiar with how rules_go works with cgo. I imagine this might cause issues on where its built... Any thoughts here?

joeljeske:
@achew22, this PR is fully ready now, with green tests! Let me know about any changes needed.

ibazel/ibazel_test.go
```diff
@@ -216,7 +212,7 @@ func TestIBazelLoop(t *testing.T) {
 	assertRun()
 	assertState(WAIT)
 	// Source file change.
-	i.sourceEventHandler.SourceFileEvents <- fsnotify.Event{Op: fsnotify.Write, Name: "/path/to/foo"}
+	go func () { i.sourceFileWatcher.Events() <- fs_watcher.Event{Op: fs_watcher.Write, Name: "/path/to/foo"} }()
```
joeljeske, line 215: Due to removing source_event_handler, i had to change this to emit async as to not block the receiving end.

ibazel/fs_watcher/fs_watcher.go
```diff
@@ -12,36 +12,23 @@
...
 import (
 	"github.com/fsnotify/fsnotify"
 )
 
-type SourceEventHandler struct {
-	SourceFileEvents  chan fsnotify.Event
-	SourceFileWatcher *fsnotify.Watcher
-}
-
-func (s *SourceEventHandler) Listen() {
```
joeljeske, line 26: This is a bad diff; this file was deleted not renamed. It is not needed since now individual files are not watched, rather just their directories

e2e/common.go
```diff
@@ -1,39 +1,22 @@
...
 var ibazelPath = getiBazelPath()
 
 func getiBazelPath() string {
-	suffix := ""
-	// Windows expects executables to end in .exe
-	if runtime.GOOS == "windows" {
-		suffix = ".exe"
+	path, ok := bazel.FindBinary("ibazel", "ibazel"); 
+	if !ok {
+		panic("Failed to locate binary //ibazel:ibazel")
```
achew22, line 17: Can you expand a bit here and say

"Failed to locate binary //ibazel:ibazel, please add it as a data dependency"?

ibazel/BUILD
achew22: This is an unfortunate thing to have to change. Turning pure off means that the binaries are no longer statically linked and we are using the cgo runtime (much slower, much buggier, much more pain) on mac.

Is this because of the dependency on fsevents?

If we decide to go this route we should definitely modify the above comment to reflect current state.

ibazel/BUILD
```diff
@@ -20,7 +20,11 @@ go_binary(
...
+    cgo = select({
+        "@io_bazel_rules_go//go/platform:darwin": True,
```
achew22, line 25: Can you explain why darwin is special here?

ibazel/BUILD
```diff
@@ -30,29 +34,35 @@ go_binary(
 go_library(
     name = "go_default_library",
     srcs = [
-        "fsnotify.go",
+        "fs_watcher_fsnotify.go",
```
achew22, line 37: Would it be possible to consolidate this logic into the `fs_watcher` package?

ibazel/BUILD
```diff
@@ -30,29 +34,35 @@ go_binary(
...
         "//ibazel/log:go_default_library",
         "//ibazel/output_runner:go_default_library",
         "//ibazel/profiler:go_default_library",
         "//ibazel/workspace_finder:go_default_library",
         "//third_party/bazel/master/src/main/protobuf/blaze_query:go_default_library",
-        "@com_github_fsnotify_fsnotify//:go_default_library",
-    ],
+    ] + select({
+        "@io_bazel_rules_go//go/platform:darwin": [
+            "//ibazel/fsevents_watcher:go_default_library"
```
achew22, line 62: Can you add a comment for future me that describe why this needs to only be included on darwin?

ibazel/fs_watcher_darwin.go
achew22: What would you think about making it opt in for a release or two and then letting it soak? I'm pretty sure that no one really beta tests this software or reads my release notes, but in case someone does it would be nice to try it out on a couple of projects before things are completely replaced.

ibazel/fs_watcher_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+// limitations under the License.
+
+// +build darwin
+
+package main
+
+import (
+	"os"
+
+	"github.com/bazelbuild/bazel-watcher/ibazel/fs_watcher"
```
achew22, line 22: What would you think about making this a part of `github.com/bazelbuild/bazel-watcher/ibazel/fswatcher` and then moving 

```
github.com/bazelbuild/bazel-watcher/ibazel/fsnotify_watcher => github.com/bazelbuild/bazel-watcher/ibazel/fswatcher/fsnotify_watcher
github.com/bazelbuild/bazel-watcher/ibazel/fsevents_watcher => github.com/bazelbuild/bazel-watcher/ibazel/fswatcher/fsevents_watcher

File: ibazel/fs_watcher_darwin.go [hunk 12]
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/bazelbuild/bazel-watcher/ibazel/fsnotify_watcher"
+	"github.com/bazelbuild/bazel-watcher/ibazel/fsevents_watcher"
+)
+
+func NewWatcher() (fs_watcher.Watcher, error) {
+	flag, ok := os.LookupEnv("IBAZEL_USE_LEGACY_WATCHER") 
+	if ok && flag != "0" {
+		return fsnotify_watcher.NewWatcher()
+	}
+	return fsevents_watcher.NewWatcher()
```
achew22, line 32: Could we wrap the event handler here to add a note to the user that says something like "You're using an experimental filesystem watcher. If you would like to disable that, please set this environment variable" if something goes wrong?

ibazel/fs_watcher_fsnotify.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !darwin
+
+package main
```
achew22, line 17: Same comment here about moving packages.

ibazel/fsevents_watcher/fsevents_watcher.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/bazelbuild/bazel-watcher/ibazel/fs_watcher"
+)
+
+type realFSEventsWatcher struct {
+	es *fsevents.EventStream
+	evs chan fs_watcher.Event
+}
+
+var _ fs_watcher.Watcher = &realFSEventsWatcher{}
```
achew22, line 30: When you move this inside `github.com/bazelbuild/bazel-watcher/ibazel/fswatcher/fseventswatcher` you'll need to remove this line due to circular imports, but that's okay, the interface will be verified by the other package.

ibazel/fsevents_watcher/fsevents_watcher.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (w *realFSEventsWatcher) Close() error {
```
achew22, line 32: Sorry the example code is bad, but would it be okay if you added a comment like

```
// Close implements ibazel/fswatcher.Watcher.
```

ibazel/ibazel.go
```diff
@@ -570,3 +544,13 @@ func parseTarget(label string) (repo string, target string) {
 	parts := strings.Split(strings.TrimPrefix(label, "@"), "//")
 	return parts[0], parts[1]
 }
+
+func keys(m map[string]struct{}) []string {
+	keys := make([]string, len(m))
+	i := 0
+	for k := range m {
+    	keys[i] = k
```
achew22, line 552: This is weirdly formatted, could you run gofmt on it?

ibazel/ibazel_test.go
achew22: This test was very carefully constructed to be able to run single threaded with a blocking channel. It's one of the few things I think I did right in the beginning on this project 😊 .

I would like to understand what you changed about the state machine that forced this to become async. Your comment implies you understand what is going on, could you speak a little bit more about it?

release/github/publish.sh
achew22: Is this actually compiling and creating new binaries in the correct output dirs? If that's the case then I think the best course is going to be for me to clone it locally and use a disassembler or something to look at how the files are actually assembled.

TL;DR: I have low confidence in the flags doing things I expect

ibazel/fsnotify_watcher/fsnotify_watcher.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
```
achew22, line 1: Big lift ask, is this testable at all?

ibazel/fs_watcher_darwin.go
joeljeske: Actually, I had originally written it where it was opt-in, but then was thinking that likely no one would ever try to use it or know it was there; and therefore we might not get *any* feedback on it. If integrated properly FSEvents should only be a positive on macOS which is why I made it a default. I figured if people encountered issues, they could create tickets for issues, and set the var to use the legacy watcher until its fixed. I'll do whatever you think is best, but was worried that we wouldn't get any results from 'letting it soak' and it would just delay turning it on for all macOS.

ibazel/BUILD
joeljeske: Oh man, I didn't realize it would change the runtime, I thought it only affected how it was compiled and assembled. The FSEvents packages does have [`C` imports](https://github.com/fsnotify/fsevents/blob/master/wrap.go#L6-L26) that link specifically with the macOS `CoreServices.framework` which includes the bindings for FSEvents. It *seemed* that I was required to make the go binary with `cgo` and `pure: off` due to this constraint, but perhaps there is another workaround I am not aware of? Any ideas?


Side note; I was unable to make `pure` configurable with a select. I think its value is used in the macro to change things. Instead of `auto` I could duplicate the `go_binary` for macOS and set `pure: off` and leave the other as `pure: on` and could configure an alias for the proper platform. Would you prefer that?

release/github/publish.sh
joeljeske: It is still compiling 3 binaries, one for each platform. I changed the bash function to accept a 3rd argument and change namings based on it. 

What I was worried about is being able to cross-compile to macOS from linux now that cgo is on. I do not know if that is possible anymore, but I *think* its a requirement to use FSEvents since its an OS framework.

ibazel/ibazel_test.go
joeljeske: So i'm new to go, so please bare with me and correct any misconceptions i may have. 

My understanding is that go channels will block and wait for the other side to be ready to receive an event. Before my change, we had [source event handler](https://github.com/bazelbuild/bazel-watcher/blob/master/ibazel/source_event_handler.go#L45) that was listening for events on a separate thread, and re-emitting events from that thread. The need for that entire file *appeared* negated since ibazel had evolved to watch parent directories, instead of files themselves. So after removing that separate thread, this channel send appeared to block as there was not a receiver ready on the other end. 

Does that sound like its correct? If so, should I remedy this issue another way besides an inline separate thread here?

ibazel/BUILD
joeljeske: Only darwin *needs* to be `cgo` since it relies on the FSEvents library which is with cgo (only macOS uses FSEvents). My understanding is that the final binary needs to be aware of cgo vs pure of all of its dependencies. Is there a way for the final binary to include cgo dependencies without being cgo itself? 

(I'll leave a comment in the code also.)

ibazel/fs_watcher_darwin.go
joeljeske: Yes, this is a great way for users to be made aware of the change and disable it if needed!

ibazel/BUILD
joeljeske: Whoops I was wrong in my understanding. We can remote this attr altogether and it works out just fine!

ibazel/fsnotify_watcher/fsnotify_watcher.go
joeljeske: Yes, I just added a test to ensure the proper calls to Add() and Remove() are performed when various UpdateAll commands are invoked. Let me know if there is other functionality you would like tests added for.

joeljeske:
@achew22, I think I incorporated all your comments. The build failure is intermittent and unrelated as far as i can tell. (I force pushed a few times to retry but still failing...) 

Let me know if there is anything else I can do!

achew22:
> @achew22, I think I incorporated all your comments. The build failure is intermittent and unrelated as far as i can tell. (I force pushed a few times to retry but still failing...)
> 
> Let me know if there is anything else I can do!

Is it always failing in the same way? Can you describe the intermittentness? The mainline branch occasionally has flakes (<5%) but only on the windows box. It's definitely strange that this is on a Linux box which I don't think has ever really been flakey

joeljeske:
Here are some build failures, always on ubuntu, but sometimes on 18.04 and sometimes on 16.04. Its related to the sentinal creation


https://buildkite.com/bazel/bazel-watcher/builds/1310#647553fb-aae1-400a-825f-041261532739
https://buildkite.com/bazel/bazel-watcher/builds/1311#d6488321-016c-4d06-90d7-b383e49b1470
https://buildkite.com/bazel/bazel-watcher/builds/1312#be1148c7-a3bc-43c0-bc96-855240c85597

There error is sometimes: 

```plain
TestOutputRunner: output_runner_test.go:97: Couldn't find sentinal. os.Stat("/tmp/fixCommandSentinel709582915"): stat /tmp/fixCommandSentinel709582915: no such file or directory
```

and sometimes it is 

```plain
TestOutputRunner: output_runner_test.go:122: Found a sentinel when expecting none: The third run should not create a sentinel.
```

Looking into the test more, I have no idea why my changes would be affecting the ibazel command runner. I have not been able to reproduce it locally, macOS or Ubuntu 20.04

joeljeske:
@achew22 its all green now.

achew22:
Can you rebase and upload again?

joeljeske:
@achew22, I have worked to fix issues in the erroring test case (see https://github.com/bazelbuild/bazel-watcher/pull/406/commits/d73d9aef6f8afaded3f3b6af4db6563b91354eab) but it is still somewhat flaky. I really would like assistance in narrowing it down. It does not seems like anything related to my code changes, but related to how the command runner is working.

achew22:
@joeljeske, I'm really busy this week and will take a look this weekend. Thanks for the ping 🙂

joeljeske:
Sorry to bother again, @achew22. Have you had any time to check this out?

achew22:
@joeljeske no problem at all. Sorry, I took a new job and it's an exciting time. Hard to find the energy for after school sports. A couple of kicks is more than fair.

Looking at the error output:

```
    TestOutputRunner: output_runner_test.go:105: ibazel invoked as: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/execroot/__main__/bazel-out/k8-fastbuild/bin/e2e/output_runner/linux_amd64_race_stripped/go_default_test.runfiles/__main__/ibazel/linux_amd64_race_stripped/ibazel --bazel_path=/usr/local/bin/bazel --log_to_file=/tmp/ibazel_output.230938439.log --run_output=true --run_output_interactive=false run //:overwrite
```

The next thing I would do is enhance the testing infrastructure (me saying this is a running joke on my team). [`output_test_runner.go:107`](https://github.com/bazelbuild/bazel-watcher/blob/d73d9aef6f8afaded3f3b6af4db6563b91354eab/e2e/output_runner/output_runner_test.go#L107) is invoking `checkNoSentinal` which could be modified to return a bool in the error case, and then you could replace the line and have it print out the ibazel output like this.

```
if !checkSentinel(t, sentinelFile, "The second run should create a sentinel.") {
  t.Error(i.GetOutput())
}
```

which would provide some very useful debugging information in this context.

All of this said, you went above and beyond in fixing that test. To me, unless you believe that test is validating something that was dramatically changed by your code, it would be totally acceptable to leave it as a TODO to fix it. Thanks for your valiant efforts in fighting the good fight!

achew22:
@joeljeske friendly ping

joeljeske:
Thanks @achew22 ill push up a change to remove that test here in a bit.

joeljeske:
@achew22, in the failing test, I suspected a race condition in the output runner e2e test since the test is unaware of when the exec runs. [I put in a simple sleep](https://github.com/bazelbuild/bazel-watcher/pull/406/commits/9d70630944aaa75080e8748f2949000faedb1ffc) and it greened the build first try. I don't like the solution, nor do I know why it started failing on my branch and not in master as it seems that it should. I still do not think that my changes *should* affect this portion of ibazel at all anyway. 

I'll defer to your judgement on next steps, whether to fix the test differently, or to proceed. Thanks for the reminder on this PR too.

e2e/output_runner/output_runner_test.go
```diff
@@ -69,9 +71,10 @@ func checkNoSentinel(t *testing.T, sentinelFile *os.File, msg string) {
 func checkSentinel(t *testing.T, sentinelFile *os.File, msg string) {
 	t.Helper()
 
+	time.Sleep(500 * time.Millisecond)
```
achew22, line 74: This `time.Sleep` totally makes sense to me because the process is supposed to be waiting for something to happen and the test is parellized. WDYT about turning this into a fast path exit:

```
deadline := time.Now().Add(5*time.Second)
var err error
for {
  if time.Now().After(deadline) {
    t.Errorf("Couldn't find sentinal. os.Stat(%q): %s\n%s\n", sentinalFileName, err, msg)
    return
  } else if _, err := os.Stat(sentinalFileName); err == nil {
    // No error stat'ing the file means it exists.
    return
  }
}
```

Now it'll tight loop until it succeeds or timeout happens.

e2e/output_runner/output_runner_test.go
```diff
@@ -69,9 +71,10 @@ func checkNoSentinel(t *testing.T, sentinelFile *os.File, msg string) {
...
 	sentinalFileName := sentinelFile.Name()
 	if _, err := os.Stat(sentinalFileName); os.IsNotExist(err) {
-		t.Errorf("Couldn't find sentinal. os.Stat(%q): %s\n%s\n", sentinalFileName, err, msg)
+			t.Errorf("Couldn't find sentinal. os.Stat(%q): %s\n%s\n", sentinalFileName, err, msg)
```
achew22, line 77: This formatting looks a little odd to me. Please remember to run `gofmt` over the repo before committing 😄

ibazel/BUILD
achew22: > Side note; I was unable to make pure configurable with a select. I think its value is used in the macro to change things. Instead of auto I could duplicate the go_binary for macOS and set pure: off and leave the other as pure: on and could configure an alias for the proper platform. Would you prefer that?

Oh `select`... your limitations are so impossible to understand...

I think I'm okay with this since auto will infer it from command line flags. It's too bad that we will have to do a cgo binary, but I think that's the price of having a good integration with osx.

ibazel/fswatcher/fswatcher.go
```diff
@@ -0,0 +1,30 @@
...
+package fswatcher
+
+import (
+	"github.com/bazelbuild/bazel-watcher/ibazel/fswatcher/common"
+)
+
+type Event = common.Event
+type Op = common.Op
+
+const Create = common.Create
```
achew22, line 24: Why type alias things that are already in the repo rather than just directly refer to them?

repositories.bzl
```diff
@@ -3,6 +3,13 @@ load("@bazel_gazelle//:deps.bzl", "go_repository")
 # bazel run //:gazelle -- update-repos -from_file=go.mod -to_macro=repositories.bzl%go_repositories
 
 def go_repositories():
+    go_repository(
+        name = "com_github_fsnotify_fsevents",
+        importpath = "github.com/fsnotify/fsevents",
+        sum = "h1:/125uxJvvoSDDBPen6yUZbil8J9ydKZnnl3TWWmvnkw=",
+        version = "v0.1.1",
+        patches = ["//:patches/fsnotify_fsevents_pr_38.diff"],
```
achew22, line 11: Can we move this file to `third_party/fsnotify/fsevents/pr_38.diff`. `third_party` is, by convention where this sort of thing would usually go.

achew22 (commented):
Sorry for the really slow response on review. This week has been hectic.

e2e/output_runner/output_runner_test.go
```diff
@@ -60,7 +61,8 @@ func TestMain(m *testing.M) {
 
 func checkNoSentinel(t *testing.T, sentinelFile *os.File, msg string) {
 	t.Helper()
-
+	
+	time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
joeljeske: Yes, you are probably correct. I was not getting this race condition locally. I mostly put it here for symmetry, since it *could* have a false positive assertion IFF the file was about to be created but was not yet created. I can remove it though since it is not needed for it to pass, only to ensure that it would not pass when it shouldnt

e2e/output_runner/output_runner_test.go
joeljeske: Yea thats a good approach, thanks for the suggestion

e2e/output_runner/output_runner_test.go
joeljeske: Yes, sorry! thanks

ibazel/fswatcher/fswatcher.go
joeljeske: Yea, I think this was a refactoring artifact. Thanks for the callout.

ibazel/ibazel_test.go
```diff
@@ -256,7 +252,7 @@ func TestIBazelLoop(t *testing.T) {
 	assertRun()
 	assertState(WAIT)
 	// Source file change.
-	i.sourceEventHandler.SourceFileEvents <- fsnotify.Event{Op: fsnotify.Write, Name: "/path/to/foo"}
+	go func() { i.sourceFileWatcher.Events() <- common.Event{Op: common.Write, Name: "/path/to/foo"} }()
```
achew22, line 255: I'm really sorry to be throwing one last thing out before we cross the finish line, but this looks very wrong to me. In this test, the state machine is single threaded with blocks in order to ensure there are no race conditions. I presume you put a state transition in a go routine because it was blocking. That means that there is an unhandled state transition, or race condition in the actual code introduced in this PR.

I went through and did a quick audit of the changes you made and they all look okay. Could you try removing the `go` I think it might just work (TM). If it doesn't can you tell me and I'll look into it with you.

ibazel/fswatcher/BUILD
```diff
@@ -0,0 +1,37 @@
...
+    srcs = [
+        "factory_darwin.go",
+        "factory_fsnotify.go",
+    ],
+    importpath = "github.com/bazelbuild/bazel-watcher/ibazel/fswatcher",
+    visibility = ["//ibazel:__subpackages__"],
+    deps = [
+        "//ibazel/fswatcher/common:go_default_library",
+        "//ibazel/fswatcher/fsnotify:go_default_library",
+    	"//ibazel/log:go_default_library",
```
achew22, line 28: Nit: the indentation is off here as well.

achew22 (changes requested):
Sorry to leave one more thing on there. I think after this we're ready to go!

ibazel/fswatcher/BUILD
joeljeske: Thanks, my .editorconfig usually handles these but this project doesn't have one. it was a tab-spacing issue.

ibazel/ibazel_test.go
joeljeske: Yea, this was an interesting one. In my refactoring, I remove "[source_event_handler](https://github.com/bazelbuild/bazel-watcher/pull/406/files#diff-58ef3a9da67919bd663ea42bf5d6b4b02516d72b70815129d241e69d135619fc).go" (git thinks it was changed, but i deleted it). In that file, there was a goroutine that watched and replayed events on the source file channel. As that was no longer needed in my refactoring, I had to put the go routine here instead or else it blocks forever waiting either for the channel sender or the first value to come through. I am very new to go, so I could be missing something else, but that is my understanding. If I remove this go func here, it hangs and times-out.

joeljeske:
```exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //ibazel:go_default_test
-----------------------------------------------------------------------------
=== RUN   TestIBazelLifecycle
[96miBazel [9:12PM][0m: You are using an experimental filesystem watcher. If you would like to disable that, please set the environment variable
	IBAZEL_USE_LEGACY_WATCHER=1
--- PASS: TestIBazelLifecycle (0.00s)
[96miBazel [9:12PM][0m: Querying for files to watch...
=== RUN   TestIBazelLoop
Error finding remote repositories directory: open external: no such file or directory
Error finding remote repositories directory: open external: no such file or directory
[96miBazel [9:12PM][0m: Demoing 
-- Test timed out at 2020-12-01 21:17:44 UTC --
```

achew22: Doesn't that mean that the order of operations for the state machine should have changed and we need to update this test to reflect the new state of the world?

joeljeske: I think the state machine operates in the same order..?
Could you explain more please?

achew22: If the state machine hasn't changed, then why would this test start failing? Something involved with the state machine must have changed, either the machine it self or the way that the inputs respond. Either way we will have to change this test, I just want to make sure that I understand what changes are being made

joeljeske: Ah ok. Yes I think before my change, the state machine was not single threaded, see the previously linked file that I removed. After my change, the state machine is single threaded but required the tests to not be single theaded.

joeljeske: Any ideas?

achew22: Sorry, I'm in a work crunch. Can I get back to you on Monday?

joeljeske: Ping! ;) Have you had any spare time for this lately @achew22?

achew22:
The more I think about this the more I think you have to do it this way. Thanks for putting up with my nonsense. Can you sync to head and fix the merge conflicts then I'll merge.

joeljeske:
@achew22 Finally we are up to date and green :) 
I'm happy with the current state of this if you are.

achew22 (approved):

achew22:
Sorry I'm such a pain in the ass! Thanks so much for your contribution. I hugely appreciate it. Can you please report back to tell me if this fixes the macos issues?

achew22:
@joeljeske I just tried putting out a release with this, but I'm running into some interesting issues:

```
ERROR: While resolving toolchains for target @io_bazel_rules_go//:cgo_context_data: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
ERROR: Analysis of target '//ibazel:ibazel' failed; build aborted: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
```

Have you seen anything like that before in your testing? I looked at the linked issue, but it doesn't look like that flag has actually been flipped. Got any ideas?

joeljeske:
I assume that is occurring when building the macOS binary, since we are now using cgo for that binary. What does the release agent look like? Are all binaries build on some container or what does that process look like? 

If all binaries are expected to be built on linux agent, then we may need to investigate configuring a macOS targeted c-toolchain on linux so that it can be used in the CGO compilation. I am not sure of the feasibility of that. 

It would definitely be easier if you could configure the binaries to be built by different agents OSs.

achew22:
The whole release process is checked into the repo [here](https://github.com/bazelbuild/bazel-watcher/blob/master/release/release.sh). I'm building on a linux box because I don't have a mac with development tools. It's built in a "normal" Ubuntu 20.04 (latest LTS fully updated) box where there are no containerization or anything like that.

> It would definitely be easier if you could configure the binaries to be built by different agents OSs.

I agree this would be easier, but I don't work at Google and don't have access to any of their build infrastructure.

joeljeske:
I do not know if I would be able to get the cross compiler toolchain setup for c/c++ so that the CGO binary can leverage it; its a bit beyond my familiarity and expertise :( 

It might be easier to setup GH Actions or similar to auto build/publish when tags/releases are created or something. That would be an improvement to your existing process anyway.

joeljeske:
Do you have any thoughts about getting past this? I feel bad for bricking your deployment process! Are you open to setting up an automated deployment builder like GH Actions or something?

achew22:
I'm extremely open to using Github actions to do releases. That would be wonderful! I spent some time reading about cross compilation, and I think that it might be as simple as loading in `rules_cc`, but I haven't gotten time for it yet

PR #410, opened 2020-07-06
Make the bug report

achew22:


achew22:
@googlebot I signed it! (already)

achew22:
@aiuto sorry to keep pinging you on this issue. Maybe I don't actually have permission to modify this file? It seems like the @googlebot and github permissions are probably not synchronized but I'm not sure.

aiuto:
I have no idea why

aiuto:
I think you can't directly modify the file because it is checked into the repo, and all PRs require a review.
At least, that's my first guess.
LMK if you can not merge this now.

PR #425, opened 2020-10-15
Add run_command_before flag

### LINKED ISSUE #423: Clear the terminal output before every run
Is there a way to clear the terminal output before every run?

I've tried an approach using `.bazel_fix_commands.json` but it clears the terminal output after running the tests.
```json
[
  {
    "regex": "go_default_test",
    "command": "clear"
  }
]
```

franciscocpg:
It adds a `run_command_before` to run a command before each execution.

The example below would run the `clear` command before each `test` execution:
```shell
ibazel -run_command_before=clear test //packages/...
```

Closes #423.


achew22:
@franciscocpg, this looks like a really great addition to the ibazel world! Thanks so much for your contribution. A couple of things that I would like to see before we merge this:

* [ ] This isn't conceptually a part of the `output_runner`, since it doesn't view the output and act on it. What would you think about making a new struct in a new package, say `github.com/bazelbuild/bazel-watcher/ibazel/lifecycle_hooks` and install it with the other lifecycle listeners?

https://github.com/bazelbuild/bazel-watcher/blob/51d9fe43bf520ce40a3c22affc960fd094fce51d/ibazel/ibazel.go#L103-L113

* [ ] It would be really great to get some test coverage over this functionality, specifically like in our [`e2e/output_runner` directory](https://github.com/bazelbuild/bazel-watcher/blob/master/e2e/output_runner/output_runner_test.go).

franciscocpg:
Hi @achew22 

Thanks for your feedback.

One question: do you think that it makes sense to put this implementation at `ibazel/live_reload/server.go` or do you prefer to create the `github.com/bazelbuild/bazel-watcher/ibazel/lifecycle_hooks` package?
https://github.com/bazelbuild/bazel-watcher/blob/51d9fe43bf520ce40a3c22affc960fd094fce51d/ibazel/live_reload/server.go#L76

achew22:
> One question: do you think that it makes sense to put this implementation at ibazel/live_reload/server.go or do you prefer to create the github.com/bazelbuild/bazel-watcher/ibazel/lifecycle_hooks package?

The `live_reload` package is mostly about sending a websocket event to the browser that informs it to reload, which I understand is a common pattern for JS/TS devs. I think this is a distinct and separate package and, frankly, one that I found myself wishing for today once you posted the idea 😁 .

If you've got any more questions please don't hesitate to ask.

franciscocpg:
@achew22 

Could you take a look if it makes sense now?

I've decided to refactor `workspace_finder` to `workspace` as well.

ibazel/lifecycle_hooks/lifecycle_hooks.go
```diff
@@ -0,0 +1,45 @@
+package lifecycle_hooks
```
achew22, line 1: Could you add a package level comment describing what this does?

ibazel/lifecycle_hooks/lifecycle_hooks.go
```diff
@@ -0,0 +1,45 @@
...
+func (l *LifecycleHooks) ChangeDetected(targets []string, changeType string, change string) {}
+
+func (l *LifecycleHooks) Cleanup() {}
+
+func (l *LifecycleHooks) BeforeCommand(targets []string, command string) {
+	if *runCommandBefore != "" {
+		commandAndArgs, err := shellwords.Parse(*runCommandBefore)
+
+		if err != nil {
+			panic(err)
```
achew22, line 37: I think it's probably reasonable to exit the program on an error here, but WDYT about using ibazel's [`log.Fatalf`](https://godoc.org/github.com/bazelbuild/bazel-watcher/ibazel/log#Fatalf) to provide a user friendly error message that also provides a few examples of what would be acceptable?

franciscocpg:
Ops, just noticed that this project has a test for Windows and the test I've added is *nix dependent. Any ideas about how to handle that?

ibazel/workspace/workspace.go
```diff
@@ -59,8 +65,35 @@ func (m *MainWorkspaceFinder) FindWorkspace() (string, error) {
 	}
 }
 
-type FakeWorkspaceFinder struct{}
+func (m *MainWorkspace) ExecuteCommand(command string, args []string) {
+	for i, arg := range args {
```
achew22, line 69: I really like this refactor out of the main module. What a good idea.

e2e/lifecycle_hooks/lifecycle_hooks_test.go
```diff
@@ -0,0 +1,72 @@
+package lifecycle_hooks
```
achew22, line 1: Out of curiosity, other than the long bit of boiler plate (I should factor that out into a reusable thing...), what did you think of writing an e2e test?

ibazel/lifecycle_hooks/lifecycle_hooks.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+
+		l.w.ExecuteCommand(commandAndArgs[0], commandAndArgs[1:])
+	}
+}
+
+func (l *LifecycleHooks) AfterCommand(targets []string, command string, success bool, output *bytes.Buffer) {
```
achew22, line 44: Would you be willing to add a flag for `run_command_after` and copy pasta the code in here as well?

achew22 (changes requested):
Looks really really good! A few small changes then we can merge and I'll do the release process with a minor version bump.

achew22:
> Ops, just noticed that this project has a test for Windows and the test I've added is *nix dependent. Any ideas about how to handle that?

I believe `bash` and `cmd` both have an `echo` command. Maybe that can be used?

e2e/lifecycle_hooks/lifecycle_hooks_test.go
franciscocpg: I liked. It was easy to understand and implement it.

ibazel/lifecycle_hooks/lifecycle_hooks.go
franciscocpg: To be honest, I'm not sure about that. This should return an error if the command is invalid in a shell. I think that should be pretty straightforward for the user to understand what's going on.

e2e/lifecycle_hooks/lifecycle_hooks_windows_test.go
```diff
@@ -0,0 +1,38 @@
...
+	bazel_testing.TestMain(m, bazel_testing.Args{
+		Main: mainFiles,
+	})
+}
+
+func TestLifecycleHooks(t *testing.T) {
+	ibazel := e2e.SetUp(t)
+
+	ibazel.RunWithAdditionalArgs("//:test", []string{
+		`-run_command_before=echo Hello`,
```
achew22, line 34: Wouldn't this work on linux as well? Why do we need to split these?

ibazel/lifecycle_hooks/lifecycle_hooks.go
achew22: That's fair, but to me, I think that an unadorned panic with stack trace is definitely not the way to go. People usually file issues against this repo when they get a stacktrace and I don't really want to field those.

franciscocpg: I see. So we can use the `log.Fatalf`. Only about the `few examples` that I think that it doesn't make sense. What do you think?

ibazel/lifecycle_hooks/lifecycle_hooks.go
achew22: Friendly ping on this

ibazel/lifecycle_hooks/lifecycle_hooks.go
achew22: A few examples might be a bit much. Totally respect that. Let's leave examples out, but write a nice error message still. Does that sound good?

achew22 (commented):
Getting really close. Just 2 more little things (and an extra test for --after) and we can merge this.

Thanks!

franciscocpg:
Hi @achew22 

I think I've tackled all review comments in 8f5d33d.

Could you take a look?

achew22 (approved):
Looks good to me. Merging as soon as CI passes

achew22:
Looks good to me! Thanks so much for taking the time to contribute and make everything better. You did a lot of cleaning, which I  greatly appreciate. I hope it wasn't too painful contributing. I'll probably be releasing a new binary for this on Friday after #422 is merged. Maybe on Monday if things at work get too hectic.

Thanks again!

franciscocpg:
Hi @achew22.

Any expected date for the release?

achew22:
Sorry about that. I thought I did it the other day, but I was wrong. I just released [v0.14.0](https://github.com/bazelbuild/bazel-watcher/releases/tag/v0.14.0) to github and NPM. If this doesn't work, please tell me!

PR #428, opened 2020-10-19
Make ibazel accept --cxxopt as a pass-through flag for bazel.

pascalfleury:
Previously, it would choke on this (and not pass the flag).


ibazel/main.go
```diff
@@ -39,6 +39,7 @@ var overrideableBazelFlags []string = []string{
 	"--config=",
 	"--copt=",
 	"--curses=no",
+        "--cxxopt",
```
achew22, line 42: Sorry to pester, but can you switch this to being a tab instead of spaces, or run `gofmt` on it?

PR #442, opened 2020-11-27
Pass custom build settings on to Bazel

FabianHenneke:
This commit adds the prefixes --// and --no// to the list of prefixes
of arguments that should be passed on to Bazel as Starlark build
settings take this form.

See: https://docs.bazel.build/versions/master/skylark/config.html#using-build-settings-on-the-command-line

### DIFF
--- a/ibazel/main.go
+++ b/ibazel/main.go
@@ -62,6 +62,10 @@ var overrideableBazelFlags []string = []string{
 	"--test_output=",
 	"--test_tag_filters=",
 	"--test_timeout=",
+	// Custom Starlark build settings
+	// https://docs.bazel.build/versions/master/skylark/config.html#using-build-settings-on-the-command-line
+	"--//",
+	"--no//",
 }
 
 var debounceDuration = flag.Duration("debounce", 100*time.Millisecond, "Debounce duration")


FabianHenneke:
Note that it is possible to define aliases for build settings which could then have arbitrary shape. In order to handle these correctly the logic concerning overrideable arguments would need to be changed more extensively, e.g. with a mechanism similar to the shell's `--`.

achew22:
I agree that parsing the command line options more properly is definitely in order. Sorry I've not had the time to make it more proper. Looks like supporting aliases will be the straw that breaks this flag parsers back.

joneshf:
Heya! Ran into this with flag aliases.

If I understand the way `ibazel`'s flag parser works, it seems like one way to support flag aliases would be to check if the string starts with a single dash. I.e. change https://github.com/bazelbuild/bazel-watcher/blob/5d163569d1ae21a5551de6abd52e08d61630a8c1/ibazel/main.go#L120-L122 to something like:
```go
func isOverrideableBazelFlag(arg string) bool {
	return strings.HasPrefix(arg, "-")
}
```
This would effectively mean that there'd be no need to explicitly state which non-startup flags are supported: https://github.com/bazelbuild/bazel-watcher/blob/5d163569d1ae21a5551de6abd52e08d61630a8c1/ibazel/main.go#L33-L71

It seems like this might work because the startup flags are checked before non-startup flags: https://github.com/bazelbuild/bazel-watcher/blob/5d163569d1ae21a5551de6abd52e08d61630a8c1/ibazel/main.go#L138-L145
so the startup flags should have precedence and be put in the right slice. But, I might be missing something else about how this all goes together.

Would that work, or is there more to this that I'm not seeing? If that would work, I can throw a PR together.

achew22:
I may be missing something with this plan, but I think at least one issue comes from this section of the [bazel documentation](https://docs.bazel.build/versions/master/guide.html#specifying-targets-to-build):

> Many Bazel commands accept a list of target patterns as arguments, and they all honor the prefix negation operator -. This can be used to subtract a set of targets from the set specified by the preceding arguments. Note that this means order matters. For example,
> 
> ```
> bazel build foo/... bar/...
> ```
> means “build all targets beneath foo and all targets beneath bar”, whereas
>
> ```
> bazel build -- foo/... -foo/bar/...
> ```
> means “build all targets beneath foo except those beneath foo/bar”. (The -- argument is required to prevent the subsequent arguments starting with - from being interpreted as additional options.)

When I was playing with this I couldn't come up with a scheme that remained consistent and valid that didn't involve effectively what we're doing today but where I wrote a program to parse the help output to get a list of all the arguments that were supported.

If you're interested in doing this, I think the best way to go about it would be to write a program that runs `bazel help flags-as-proto` and parses the output as a [`FlagCollection `](https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/bazel_flags.proto) and then emits the list of flags into a `.go` source file. That would be a pretty simple task.  Thus, solving the problem once and for all.

once and for all

joneshf:
Oh that's neat! I didn't realize you could negate targets. Thanks for explaining that!

> The `--` argument is required to prevent the subsequent arguments starting with `-` from being interpreted as additional options.

So, this seems like negated targets currently don't work in `ibazel` since it throws everything after the `--`: https://github.com/bazelbuild/bazel-watcher/blob/5d163569d1ae21a5551de6abd52e08d61630a8c1/ibazel/main.go#L131-L135 to the underlying command, not as targets to `bazel`: https://github.com/bazelbuild/bazel-watcher/blob/5d163569d1ae21a5551de6abd52e08d61630a8c1/ibazel/main.go#L127-L130 Is that right? So the idea here is to fix the current parser while also adding support for flag aliases?

> If you're interested in doing this, I think the best way to go about it would be to write a program that runs `bazel help flags-as-proto` and parses the output as a base64 encoded [`FlagCollection`](https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/bazel_flags.proto) and then emits the list of flags into a `.go` source file.

I'd be down to give it a shot.

Didn't know there was a `flags-as-proto` subcommand. I couldn't find any documentation about it. All I could find was the code around it: https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java#L157-L159. Do you have any more information about it?

The reason I ask is because I can't figure out how to get it to pick up either custom build setting or flag aliases. So, I'm wondering if there's some other flag that has to be passed to it or something. If not, I'm not sure how using `bazel help flags-as-proto` helps with knowing which flags passed to `ibazel` are custom build settings or flag aliases. I.e. it seems like there's still going to have to be custom logic to parse both of these things with or without the `bazel help flags-as-proto`.

PR #446, opened 2020-12-09
Simplify setting of execute bit

smklein:
Instead of providing multiple avenues for making shell scripts executable, converge on "e2e.SetUp" internally invoking directory walk + chmodding

### DIFF
--- a/e2e/common.go
+++ b/e2e/common.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/bazelbuild/rules_go/go/tools/bazel"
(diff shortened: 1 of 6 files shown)


e2e/common.go
```diff
@@ -77,8 +78,10 @@ func SetExecuteBit(t *testing.T) {
 			return err
 		}
 
-		if err := os.Chmod(path, 0777); err != nil {
```
smklein, line 80: It seems this was previously setting *everything* to be executable, not just shell scripts!

achew22: Oops, good catch!

achew22 (approved):

achew22:
Thanks so much for not just contributing, but improving the e2e testing framework!

PR #452, opened 2021-01-13
Add --compile_one_dependency as overrideable flag

akonradi:
Documented here: https://docs.bazel.build/versions/3.7.0/user-manual.html#flag--compile_one_dependency

### DIFF
--- a/ibazel/main.go
+++ b/ibazel/main.go
@@ -37,6 +37,7 @@ var overrideableBazelFlags []string = []string{
 	"--build_tag_filters=",
 	"--build_tests_only",
 	"--compilation_mode",
+	"--compile_one_dependency",
 	"--config=",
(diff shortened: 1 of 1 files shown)


achew22:
What an interesting flag! Thanks for sending in the PR.

akonradi:
I'm trying this locally and I don't know that it actually does the right thing. It keeps Bazel from barfing while querying for targets, but I think the list of watched files is incomplete because changing them doesn't cause ibazel to run another build.

achew22:
I wrote the code a long time ago, but I don't think flags passed into `ibazel` are passed into the `bazel query`. This would only affect the flags sent to the build, test or run invocation.

Or, I'm totally wrong and forgot how it worked

PR #456, opened 2021-01-25
GH Actions release binary builders

joeljeske:
@achew22, converting most of the release flow to GH Actions as macOS hardware is seemingly required for cgo on macOS. 

per: https://github.com/bazelbuild/bazel-watcher/pull/406

I *think* this is a truly compatible release flow. Biggest changes:

- You are still creating release notes, tag (and pushing the tag) from your local (./release/release.sh)
- The binaries are built only once and shared between GH and NPM (windows + linux from ubuntu, and darwin from macOS)
- If the build(s) fail, your local release script does not detect it and rollback (not sure how you want to handle this one)
- You need to add "NPM_TOKEN" as a GH Secret to allow publishing to NPM from an action (https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow)

Let me know what you think, of course this could be done many different ways, but I took an approach that seemed to be close to what you were already doing.


joeljeske:
For what its worth, I was going to add a side-by-side pre-submit check in GH Actions, but running them on macOS may be to costly, given that GH allots only 2000 min/month for all exec, and for macOS they charges 10x.

achew22:
@joeljeske Sorry this is taking a while to look at. I really feel like I need to think about it before committing since I don't want to become the next supply chain attack 😉 . I'm starting to feel like I know what's going on after reading a lot of GH actions documentation, but I'm not quite there yet

.github/workflows/release.yaml
```diff
@@ -0,0 +1,108 @@
...
+          cp "release/npm/index.js" "npm-staging/index.js"
+          bazel run --config=release "//release/npm" -- "${PWD}/CONTRIBUTORS" > "npm-staging/package.json"
+
+      - name: Publish NPM Package
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          echo -n "Publishing to NPM as "
+          grep "version" < "npm-staging/package.json"
+          cd "npm-staging" && find . && npm publish
```
achew22, line 108: My biggest question is what happens right here. I have 2FA set up on my NPM account. How will this handle that?

joeljeske: You would store the "automation token" as a GH secret as "NPM_TOKEN" for use in this step. Npm has the concept of automation tokens that bypass 2fa (https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#continuous-deployment)

I personally do not have experience with it, but appears that this is the suggested flow.

joeljeske:
@achew22, no worries at all! I appreciate that you are dilligent in reviewing changes, especially ones like this that change things up a bit. I would have no hurt feelings if you felt more comfortable with a different approach (e.g. archiving the npm package bundle and then manually publishing from your own machine).

.github/workflows/release.yaml
joeljeske: And.... it looks like that "automation token" link is dead on npmjs ha, sorry.

It appears you can create a new token from going to your profile on the top-right -> access tokens -> new access tokens -> automation.

achew22:
Okay, let's see what happens! Thanks so much @joeljeske I hugely appreciate all your efforts here. This is much better than what I had been doing and is a great lanchpad for future release enhancements.

achew22:
@joeljeske this was the easiest release I've done yet. Thanks so much! It feels like we're very close. 

```
npm ERR! code E404
npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@bazel%2fibazel - Not found
npm ERR! 404 
npm ERR! 404  '@bazel/ibazel@0.15.0' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

https://github.com/bazelbuild/bazel-watcher/actions/runs/542055506

Do you know if it's possible to trigger it without actually hitting the release button and causing a new tag?

joeljeske:
>It feels like we're very close.
@achew22 hmmm... Did you remember to generate a NPM token and add it to github secrets? 

It should be automatically picked up by npm and interopolated into the [.npmrc](https://github.com/bazelbuild/bazel-watcher/blob/master/.npmrc) It should be just a hex string token. You would need to have created 

https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#continuous-deployment


> Do you know if it's possible to trigger it without actually hitting the release button and causing a new tag?

You can create a release draft to try it out i believe. You would not have to run your scripts locally, just create a release draft in GH UI. Currently the pipeline relies on information about the release being available so it knows what to modify. WE could create a different *similar* pipeline that can be run on-demand or on other triggers.

achew22:
> > It feels like we're very close.
> > @achew22 hmmm... Did you remember to generate a NPM token and add it to github secrets?

Yes, sir.

Screen Shot 2021-02-08 at 10 26 28 AM


> It should be automatically picked up by npm and interopolated into the [.npmrc](https://github.com/bazelbuild/bazel-watcher/blob/master/.npmrc) It should be just a hex string token. You would need to have created
> 
> https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#continuous-deployment
> 
> > Do you know if it's possible to trigger it without actually hitting the release button and causing a new tag?
> 
> ~You can create a release draft to try it out i believe. You would not have to run your scripts locally, just create a release draft in GH UI.~ [Edit, I read [the docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release) more carefully which states drafts are ignored.] Currently the pipeline relies on information about the release being available so it knows what to modify. We could create a different _similar_ pipeline that can be run on-demand or on other triggers.

Lemme try the draft release thing and see what happens.

achew22:
I just created a [draft release](https://github.com/bazelbuild/bazel-watcher/releases/tag/untagged-41451b929853b4606bfe) and it looks like it isn't going to be run by the releaser script until I publish it, which I'm a little reluctant to do.

achew22:
Also, it looks like that release is only visible to maintainers on the repo :-(

joeljeske:
Created a new PR to try to fix the issues
https://github.com/bazelbuild/bazel-watcher/pull/459

PR #459, opened 2021-02-09
Update release flow to be invoked manually

joeljeske:
Hey @achew22, sorry for the runaround here. I changed up the workflow so it can be run manually, in which case it does not upload the binaries to the github release nor does it publish to npm. It does, however, archive the 3 binaries and the npm package to the GH Action run result so you can download and inspect it. 

Here is a sample run from my fork:
https://github.com/joeljeske/bazel-watcher/actions/runs/551749480
 
I *think* I have fixed the npm upload issue also, which may have been due to the `.npmrc` file in the parent directory and not in the enclosed npm package, however, I still cannot fully test this publishing.

Let me know your thoughts!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,6 +3,7 @@ name: Build Release Binaries
 on:
   release:
     types: [created]
+  workflow_dispatch: {}
     
 jobs:
   build: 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -72,37 +77,41 @@ jobs:
...
 
       - name: Install Bazelisk
         run: npm --global install @bazel/bazelisk@latest
 
+      - name: Prepare ibazel binaries
+        working-directory: ./npm-staging/bin
+        run: |
+          mv darwin_amd64/ibazel_darwin_amd64 darwin_amd64/ibazel
+          mv linux_amd64/ibazel_linux_amd64 linux_amd64/ibazel
+          mv windows_amd64/ibazel_windows_amd64.exe windows_amd64/ibazel.exe
```
joeljeske, line 93: I had to re-arrange the npm package as it varied slightly from previous releases and I did not want to break the structure for folks.

achew22:
Let's try it. Thanks for being persistent!

achew22:
https://github.com/bazelbuild/bazel-watcher/actions/runs/553125824

joeljeske:
Not sure what to make of that, bazel failed to startup it seems. Haven't seen that before. Try again?

achew22:
Retriggered https://github.com/bazelbuild/bazel-watcher/actions/runs/553125824

joeljeske:
So that looks like it's failing due the go package not finding the version from the git tag. That's to be expected if there is not a tag. You could either make a RC tag or modify the go package to run without a version. Thoughts?

joeljeske:
@achew22, friendly ping here: 

If you want, we can change the behavior of the workspace status to support being invoked if there is not a current tag. Otherwise, I think that it is OK that this most recent failure occurred since it was run on a commit without a tag. 

If you are ok, you could try pushing a new release, or you could make a RC tag or something and try it out manually again. 

Thoughts?

achew22:
@joeljeske I promise I haven't forgotten about this. It's been two long weekends back to back for me. I'll probably take a look at this on Monday

joeljeske:
No worries at all, more so I just feel compelled to un-break your stable branch 😅

PR #462, opened 2021-03-04
bazel: simplify done channel rendezvous

matttproud:
Simplify the channel rendezvous to instead of using a buffered channel
with an unergonomical `ch <- struct{}{}` use a `defer close(ch)`
semantic.  This change is within its scope of usage isomorphic.

### DIFF
--- a/bazel/bazel.go
+++ b/bazel/bazel.go
@@ -233,10 +233,8 @@ func (b *bazel) Info() (map[string]string, error) {
 	stdoutBuffer, _ := b.newCommand("info")
 
 	// This gofunction only prints if 'bazel info' takes longer than 8 seconds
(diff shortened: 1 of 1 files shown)


achew22:
Thanks for sending in the PR! Out of curiosity, why were you reviewing this section of the code?

matttproud:
Long story short: I found https://github.com/derivita/bazelpackagesdriver, because I had been interested in https://github.com/golang/go/issues/34341.

I did a code study of the implementation of the driver, which sent me here next.  I figured that it was using a subprocess to manage interfacing with Bazel, which I confirmed.  Was somewhat interested if any creative solution to making the data access more scaleable was found (mediating driver model with Bazel).

I figured some of this information may be relevant for adapting this to Basel’s big brother.

Issue #463, opened 2021-03-05
iBazel rebuild fails on Windows 


lukasholzer:
**Describe the bug**
When running a bazel runfile with `ibazel` and having the `"ibazel_notify_changes"` tag set on the binary a change at a dependency is failing the build.

**Reproduction instructions**

Steps to reproduce the behaviour:
1. `https://github.com/lukasholzer/ibazel-windows-bug`
2. `ibazel run godev` (same happens when running `ibazel run nodedev` – it is not related to the binary impelementation)
3. Edit `main.js` with the test
4. See the error from the screenshot below in your terminal

**Expected behaviour**
When running the dev server with ibazel it should not crash once a file is changed.

**Version (please complete the following information):**
 - OS: **Windows 10**
 - Browser  **(does not matter)**
 - ibazel Version **`@bazel/ibazel@0.14.0` (installed globally)**
 - Bazel version
   ```
    Build label: 4.0.0
    Build target: bazel-out/x64_windows-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar    
    Build time: Thu Jan 21 07:39:16 2021 (1611214756)
    Build timestamp: 1611214756
    Build timestamp as int: 1611214756
   ```
 
**Additional context**

![image](https://user-images.githubusercontent.com/11156362/110098444-7f5d6a80-7da0-11eb-85a9-d3641dc79811.png)

achew22:
Based on the wave of thumbs on this, I'm going to guess this is a pretty common thing. Unfortunately I don't have any windows boxes I can test anything on. I do have tests running on Windows during CI and a very nice e2e testing framework that makes it pretty simple to create a testcase like this. Is there a way for someone from the thumbs up group (@thomaspink, @philippraab, @mikekuss, @nimrod13, @robaxelsen, @ffriedl89, or @rowa-audil) could add an e2e test that fails?

You could duplicate this function:

https://github.com/bazelbuild/bazel-watcher/blob/bf4736b09705671c6a841f02b9020dee40684c96/e2e/simple/simple_test.go#L65-L71

and rename it to `TestWindowsBuildLockIssue463` and have the `ibazel` object run these steps in a test. If you need more files in the synthetic repo, these are examples of how to create more files in the synthetic repo:

https://github.com/bazelbuild/bazel-watcher/blob/bf4736b09705671c6a841f02b9020dee40684c96/e2e/simple/simple_test.go#L15-L57

PR #464, opened 2021-03-09
tests(notify_changes): Add test scenario for notify changes.

lukasholzer:
This should demostrate the bug of the issue #463


achew22 (changes requested):
@lukassholzer, things are looking good but it looks like the test is back to passing which... I guess is a 😦 ? Any chance you can keep poking at it to reproduce?

PR #471, opened 2021-04-12
chore(NA): adds --show_result into the iBazel proxied flags list

mistic:
I've just found `--show_result` as an useful flag to be proxied into bazel and that should be added into that list


mistic:
@achew22 do you think this is okay to go with?

achew22:
So sorry about this you got lost in the thousand tab purge that happened yesterday.

I'll also do a release with this included

achew22 (approved):

mistic:
Thank you very much @achew22 ❤️

Issue #358, opened 2020-03-12
How do you enable JS serving?


sgammon:
I've noticed libraries like `rules_closure` are able to trigger the live-reload server to activate, such that it serves the JS code to enable live-reloading at a local port (`http://localhost:35729/livereload.js`, for example), and so on. But how does one opt-in to this behavior?

I've affixed `ibazel_live_reload` and `iblaze_notify_changes` to my binary target, but to no avail. The endpoint still refuses connection for me. Am I missing a step?

sgammon:
@achew22 / @dougkoch would you possibly know how to address this? thank you in advance for any help you are able to offer

achew22:
Did you add those as a tag on the target? Can you create an example repo that demonstrates this failure?

sgammon:
@achew22 i added the tags to a java_binary target, yeah. I’ll try to dig in and create a repro, but I’m not entirely sure it’s a failure vs. user error

if all it should take is the tags, then I suppose this is a bug?

achew22:
It should only take the tags, and I use this on a daily basis through `ts_devserver` so I don't believe it's broken.

It's a bug only if you can provide a reproduction that I can observe and add a test case for. Otherwise I'm just guessing in the dark at something that might be user error. I don't work on this professionally so I can't provide the service you're asking for. I'm sorry.

sgammon:
@achew22 no worries, i honestly appreciate the effort nonetheless, especially if you don't manage the repo for Google! lol

sgammon:
@achew22 i'll give it a try in an isolated repro, or perhaps dig into how `ts_devserver` uses it and see if i can spot any differences. thank you, will report back.

achew22:
@sgammon, I wanted to follow up here and point you at an example server https://github.com/bazelbuild/bazel-watcher/pull/378. It works great, but I can't quite a different e2e test that I wrote to work with it in Windows. Buyer beware, but hopefully that'll get you off to the races on how to parse the STDIN signaling.

I'm going to close this, if you have any more issues please, please file another issue or reopen this one.

sgammon:
@achew22 apologies but i'm still unable to grab the `IBAZEL_LIVERELOAD_URL` variable.

for context, i'm running my app this way:
`Makefile` > `ibazel` > `bazelisk` > `bazel` > `java_binary` wrapper script > `jar`

[...]

PR #473, opened 2021-04-25
Add flag to force-enable JS serving

sgammon:
For unclear reasons, `ibazel` doesn't detect our use of the `ibazel_live_reload` tag, and therefore never starts the live reload server. Fixes and closes bazelbuild/bazel-watcher#358.

When our server loads up and runs, it never sees the environment variable. If we try to load the `lrserver`'s `livereload.js` at the default port, it fails -- because the server never starts listening in the first place.

This adds a simple flag to `ibazel`: `force_live_reload`. When this flag is passed, `ibazel` starts up the `lrserver` in `Initialize` instead of during target scanning.

Changes enclosed:
- [x] Add `force_live_reload` flag, init `lrserver` in `Initialize` when passed
- [x] Elide scan for `ibazel_live_reload` tag when `-force_live_reload` is active (the server would already be on)

### DIFF
--- a/ibazel/live_reload/server.go
+++ b/ibazel/live_reload/server.go
@@ -32,6 +32,8 @@ import (
 
 var noLiveReload = flag.Bool("nolive_reload", false, "Disable JavaScript live reload support")
 
+var forceLiveReload = flag.Bool("force_live_reload", false, "Force-enable JavaScript live reload support")
+
 type LiveReloadServer struct {
 	lrserver       *lrserver.Server
 	eventListeners []Events
@@ -47,7 +49,12 @@ func (l *LiveReloadServer) AddEventsListener(listener Events) {
 	l.eventListeners = append(l.eventListeners, listener)
 }
 
-func (l *LiveReloadServer) Initialize(info *map[string]string) {}
+func (l *LiveReloadServer) Initialize(info *map[string]string) {
+	if *forceLiveReload {
+		log.Log("Force-activating live reload.")
+		l.startLiveReloadServer()
+	}
+}
 
 func (l *LiveReloadServer) Cleanup() {
 	if l.lrserver != nil {
@@ -56,15 +63,17 @@ func (l *LiveReloadServer) Cleanup() {
 }
 
 func (l *LiveReloadServer) TargetDecider(rule *blaze_query.Rule) {
-	for _, attr := range rule.Attribute {
-		if *attr.Name == "tags" && *attr.Type == blaze_query.Attribute_STRING_LIST {
-			if contains(attr.StringListValue, "ibazel_live_reload") {
-				if *noLiveReload {
-					log.Log("Target requests live_reload but liveReload has been disabled with the -nolive_reload flag.")
+	if !*forceLiveReload {
+		for _, attr := range rule.Attribute {
+			if *attr.Name == "tags" && *attr.Type == blaze_query.Attribute_STRING_LIST {
+				if contains(attr.StringListValue, "ibazel_live_reload") {
+					if *noLiveReload {
+						log.Log("Target requests live_reload but liveReload has been disabled with the -nolive_reload flag.")
+						return
+					}
+					l.startLiveReloadServer()
 					return
 				}
-				l.startLiveReloadServer()
-				return
 			}
 		}
 	}


achew22:
@sgammon, thanks for the PR! I'm definitely interested in making js live reload work for everyone!

Before we get too far down this road, let me describe my concerns. I'm concerned that if we adopt this approach, ibazel will move toward a thing that I've long been afraid of, "another app with an incompatible set of command line arguments that I can never remember". The value in having params encoded in `BUILD` files is remarkable from that perspective. Additionally, I'm concerned that there is something wrong in the live reload detection logic and I would REALLY like to figure that out since it's probably driving other people crazy.

As I understand it, you've got a `tag` set correctly on a target in your build graph. Is there any way you can make a runnable example that's available on github that exhibits this problem where it has the tag but the tag isn't detected by `ibazel`?

sgammon:
hey there @achew22 - i agree, and i would be happy to diagnose it :) for now, though, we have deadlines and not having live-reload nerfs our use of `ibazel` pretty hard. we see serious slowness with the `bazel info` phase, too (it takes 3-5 mins to start the server), so we're hoping to just get this out of the way for ourselves and fix that, and then use the productivity gains to work back upstream and get things fixed/merged.

in the meantime, i've found the source of my earlier confusion about the env var showing up. the env var shows up fine, it's just that the tag is never detected, and so the server never starts. by adding this flag, the server starts fine, env var is there, server loads, all is well with the world.

our app is based on a homegrown OSS framework ([elide](https://github.com/sgammon/elide)), and if it applies to us, it likely applies to any targets wired by macros defined there. so i suspect i can make a minimal repro, and drop that here soon.

achew22:
> for now, though, we have deadlines and not having live-reload nerfs our use of ibazel pretty hard.

Good old deadline pressure. Sorry to hear that!

> we see serious slowness with the bazel info phase, too (it takes 3-5 mins to start the server), so we're hoping to just get this out of the way for ourselves and fix that, and then use the productivity gains to work back upstream and get things fixed/merged.

That comment about the `bazel info` phase is really concerning to me WRT the configuration of your repo. From my experience at Google with Blaze and a [+2bn loc](https://www.wired.com/2015/09/google-2-billion-lines-codeand-one-place/) repo I got `bazel info` results back in much less than 5s.

> in the meantime, i've found the source of my earlier confusion about the env var showing up. the env var shows up fine, it's just that the tag is never detected, and so the server never starts. by adding this flag, the server starts fine, env var is there, server loads, all is well with the world.

Good to hear that this is alleviating the pressure. 

> our app is based on a homegrown OSS framework (elide), and if it applies to us, it likely applies to any targets wired by macros defined there. so i suspect i can make a minimal repro, and drop that here soon.

As a datapoint, at my work we have a fairly elaborate macro based framework for our servers and do not suffer from this.

I don't want you don't feel like I'm goalpost moving, so let me explicitly state my position. I'm 90% confident that I'm not going to merge this since it feels like we're hacking around either a bug in ibazel or a misconfiguration of your repo. If your launch pressure is strong enough, a strategy for advancing might be to build your own ibazel binary and distribute it with these changes. I would like to keep the conversation going so we can figure out if it is an ibazel or configuration problem though. A minimal repro would be really great for figuring this out.

sgammon:
@achew22 forgive me, perhaps that deadline pressure is leaking through in my commentary, because i see your position, understand, and agree 😄

when i first submitted, i felt like this PR was a good complement to `nolive_reload`, but now i realize that is meant for suppression of the live reload feature _without_ requiring changes in code. so that does make sense and explains why a flag the other way might be different.

i also understand and agree with your point about there likely being a bug here that needs fixing, which should get fixed, rather than papering that behind a flag. so let me rephrase! lol

we _are_ distributing this, with the fix, as a local binary and using it for our needs. that's fine for now, our team is relatively small. i would love to put together a repro that forces this out into the open.

> Good old deadline pressure. Sorry to hear that!

thank you, beauty is painful or at least that's what they tell me

> That comment about the `bazel info` phase is really concerning to me WRT the configuration of your repo. From my experience at Google with Blaze and a +2bn loc repo I got bazel info results back in much less than 5s.

wow, that is really concerning. okay. it sounds to me like we're doing something absurd somewhere along the way. would the profiling tools built into `ibazel` help diagnose this, or would the bazel target graph be the first thing you'd reach for?

> As a datapoint, at my work we have a fairly elaborate macro based framework for our servers and do not suffer from this.

that is comforting and thank you for sharing. i suspect our use of macros is rather naive, i wired together the pieces of this framework while learning bazel's internals, which admittedly still confuse me greatly. intuitively, i figure this is related to flags in a `.bazelrc` somewhere, or those macros. where would you begin diagnosing, if different from the previous answer?

thank you, by the way, for your help on this. i'll get this repro up as soon as i can.

Issue #478, opened 2021-07-16
Aborting an in progress build when a new one is needed


mistic:
I believe this is more a suggestion to improve `iBazel` then a bug report. 

I think it will be very useful to abort an in progress build when we got the information a new one is needed. For instance, I had `iBazel` running, switched to master, and then pulled upstream. Unfortunately `iBazel` started running the build as soon as I switched branches which means it was running a bunch of builds which were immediately invalidated by the pull from upstream so I needed to kill `iBazel` and restart it to avoid waiting an extra 100+sec for it to rebuild a ton.

What do you think about this @achew22 ?

achew22:
Hey @mistic,

Just to sketch out how I would think about this:

In my experience, things that are controlled through environment variables have very poor discoverability unless they're required and cause error output if they aren't provided so I think I'd like to stay away from that. My preference would be to make it similar to `ibazel_skip_live_reload` et al. that way it can be done on a per-target basis in the build graph without the need to specify on a regular basis.

You'd need to augment

https://github.com/bazelbuild/bazel-watcher/blob/4683e4f3788bb3312ee82dbdf937aa1c459c0de9/ibazel/ibazel.go#L391-L393

to track your new `ibazel_abort_compilation_early` (not suggesting a name, just needed something in there as a placeholder).

If I understand things correctly, what you'd like to do is modify the state machine to have an additional transition from the RUN state that is able to short circuit the test/build process, correct?

To write out a rough outline of this, I think you'd be extending this code to do a select on a set of channels:

https://github.com/bazelbuild/bazel-watcher/blob/4683e4f3788bb3312ee82dbdf937aa1c459c0de9/ibazel/ibazel.go#L324-L330

where the `select` is switching for events either from a i.sourceFileWatcher.Events() or to the completion of the command invoked in `RUN`. The way I'd expect the `done` to happen would be to fork off a go routine in the RUN state that is feeding `done := chan struct{}` when it's ready with `done <- struct{}{}`. Then you'd put the select below the go routine kickoff looking at both the channel and the sourceFileWatcher event feed (and maybe also the buildFileWatcher).

mistic:
[...]

PR #479, opened 2021-08-06
chore(NA): add support for abort compilation early

mistic:
Closes https://github.com/bazelbuild/bazel-watcher/issues/478

This is my first take to solve this problem and my very first bits on GO. I think it's almost there other than a single problem: I think the cancelation from the previous build,run,test is not working as expected as the new commands still got queued in the Bazel server before the old ones finishes and everything endup to eventually run and log out.

The erroneous output looks like the following and the `Another command` messages keep increasing if you keep editing while builds are in backgroud:

```
iBazel [1:36PM]: Building //packages:build
iBazel [1:36PM]: Changed: "FILE_PATH". Rebuilding...
iBazel [1:36PM]: Changed: "FILE_PATH". Canceling previously and rebuilding....
iBazel [1:36PM]: Building //packages:build
iBazel [1:36PM]: Build error: signal: killed
iBazel [1:36PM]: Building //packages:build
Another command (pid=10833) is running. Waiting for it to complete on the server (server_pid=9158)...
Another command (pid=10833) is running. Waiting for it to complete on the server (server_pid=9158)...
```

@achew22 could you take a look and help me figuring this out? I would really like to be able to move this forward. I think the problem has to do with the way the go routine and the cancelation mechanism is implemented as I think the current code running in the go routine is blocking and as such will queue up and endup executing. Also I believe we don't have a current proper way of instant killing the bazel `build,test` commands running. I would say we need to create a non blocking interface to run `build,test` commands with a proper mechanism to instantly kill the bazel commands without no further output or execution time, and then call that new function in the go routine. Could you give me some help here?


ibazel/ibazel.go
```diff
@@ -322,15 +329,75 @@ func (i *IBazel) iteration(command string, commandToRun runnableCommand, targets
...
+			return
+		}
+
+		// use default compilation behaviour
+		<-runDone
+
 	}
 }
 
+func (i *IBazel) logRunAbort(errorName string, abortType string) {
```
achew22, line 375: Once you make another state I think this logic will be better folded into those two states.

ibazel/ibazel.go
```diff
@@ -340,8 +407,19 @@ func verb(s string) string {
...
+	if i.b != nil {
+		i.b.Cancel()
+	}
+
+	if i.cmd != nil {
+		i.cmd.Kill()
+	}
+}
+
 func (i *IBazel) build(targets ...string) (*bytes.Buffer, error) {
```
achew22, line 420: Break this out into 2 functions `buildWithEarlyAbort` and just `build` and then inline the logic here. I'd like to avoid storing `i.b` if at all possible. That's too much state to be managing.

achew22 (changes requested):
This looks like a great start! Thanks so much for your work so far.

Next up will be adding some e2e tests to validate the behavior doesn't accidentally regress in the future. Here is an [example](https://github.com/bazelbuild/bazel-watcher/blob/master/e2e/lifecycle_hooks/lifecycle_hooks_test.go). The only thing I think you need to be aware of is that the `e2e` package is pretty clever about reading `stdout` and things before it advances the test so you may need to put some `sleep` statements in your `mainFiles` so that there is enough latency to guarantee that your test is actually testing what you want it to.

achew22:
> @achew22 could you take a look and help me figuring this out? I would really like to be able to move this forward. I think the problem has to do with the way the go routine and the cancelation mechanism is implemented as I think the current code running in the go routine is blocking and as such will queue up and endup executing. Also I believe we don't have a current proper way of instant killing the bazel build,test commands running. I would say we need to create a non blocking interface to run build,test commands with a proper mechanism to instantly kill the bazel commands without no further output or execution time, and then call that new function in the go routine. Could you give me some help here?

I provided some suggestions in the review that I think will help with this. Right now you're tracking `i.b` as a Bazel object outside the invocation loop. This means that it is pretty easy to accidentally leak a subprocess by overwriting your reference to it (that's what I think is causing your issue here). By moving the management of the Bazel object back inside the `run` function et al, you'll be able to scope the risk of leaking the subprocess to just that function, which will be a lot easier to review (20 lines of code to audit instead of many thousands).

On the off chance you're not familiar with it, Go has a nice feature called `defer` which you may find useful specifically in this case. Defer is a keyword that takes a function that's being called and invokes it (in a reverse stack order if you call it multiple times) at the exit of a function. It's often used for cleaning up. So, at the top of your async `build`, you could defer cancelling or killing the subprocess.

Also, just as an additional datapoint, you can't consider `bazel` to be dead because you sent it a `SIGINT`. You need to wait for it to actually terminate. If you send a `SIGKILL` to it you will kill the Bazel server in addition to the Bazel client which will wipe away your analysis cache and cause your next build to be a couple of orders of magnitude longer which will probably completely wipe away the gains you got from early aborting the build process.

mistic:
@achew22 thanks for the feedback here. I've pushed some more work into this one in order to move along with the ideas and the feedback you left. At this point I'm only interested into make sure the cancel/rebuild logic is working for the build method and not to much worried with the other methods (thats why I'm explicitly calling `i.buildWithEarlyAbort` in the run loop) as I will address that part later. However the current implementation is still stacking executions, I really believe the cancel is not working as expected. Could you provide some help here? Will we need to adapt your `command.go` in order to make it able to run multiple targets and also use it for `build` and `test` commands instead of using the go context and exec wrapping alternative for those?

mistic:
@achew22 I think I've made some changes according with your feedback. I've also add some more work (but only concerning the build step yet, thats why I'm calling inline the `i.buildWithEarlyAbort` in the state machine, code organization will go next). You know the codebase much better than me, and also as I said before I hadn't any Golang experience prior to this, so how do you think we should organize the rest of the implementation? I'm thinking about moving the logic into the `ibazel.go` file.

mistic:
@achew22 I've made some changes according with your feedback. I've also added some more work (but only concerning the build step yet, for run and test I only have placeholder functions for now). You know the codebase much better than me, and also as I said before I hadn't any Golang experience prior to this, so a couple of questions:

1 - Do you think the current code organisation makes sense according the fine sense you have around the codebase?

2 - I'm sending an inline SIGTERM to cancel the commands, do you think we should abstract this with a better and safer logic as you did in for the default command implementation?

Finally also want to mention that I've added a couple of e2e test cases (only for the build method at the moment)

mistic:
@achew22 from my previous comment https://github.com/bazelbuild/bazel-watcher/pull/479#issuecomment-896983051 I've added support for the `run` command so the `point 2` I think is solved (the current behaviour implemented for run is already compatible with the early abortion behaviour so I've just called the old method). 

I believe the only thing missing on this PR is what I describe in the previous comment `point 3`: implement a better early cancel on Windows systems. The current behaviour will just cancel the context which is super slow and not as performant as implementing a simple SIGTERM on unix based systems. Do you remember any performant way of doing this on Windows? (maybe a way to send a simulated ctrl break into the process? Or maybe just use the same logic of using a process group and terminate it entirely on Windows? ) 

If we solve this Windows part I think the PR is good to go.

Also while testing this on Windows I endup finding another problem and opening https://github.com/bazelbuild/bazel-watcher/pull/480

mistic:
@achew22 This PR is now ready for review. I've successfully implemented the cancelable behaviour across macOS, Linux and Windows. On macOS and Linux I'm just sending a `SIGTERM` which kills the entire process group while on Windows I'm sending a `CTRL+C` event (which works well and is far easier than using Job Objects or trying to reuse/refactor the logic you are using for the run command around Job Objects). I've also added E2E tests for that feature as well as spent some time manually testing the feature across the mentioned 3 operating systems. I really think we are good to go with that PR 🎉 **Just one note:** during this work I found a bug on Windows with iBazel which is fixed on another PR I've opened https://github.com/bazelbuild/bazel-watcher/pull/480. In order to correctly finish the features on this PR I had the need to cherry-pick https://github.com/bazelbuild/bazel-watcher/pull/480 here so before merging and go ahead with this current PR we should merge https://github.com/bazelbuild/bazel-watcher/pull/480.

ibazel/ibazel.go
```diff
@@ -250,23 +268,26 @@ func (i *IBazel) setup() error {
...
 	i.args = args
-	return i.loop("run", i.run, []string{target})
+	cmd := RunnableCommand{command: i.run, cancelableCommand: i.runWithEarlyAbort}
+	return i.loop("run", cmd, []string{target})
 }
 
 // Build the specified targets in the IBazel loop.
 func (i *IBazel) Build(targets ...string) error {
-	return i.loop("build", i.build, targets)
+	cmd := RunnableCommand{command: i.build, cancelableCommand: i.buildWithEarlyAbort}
```
mrmeku, line 277: How does the user get to choose between early abort vs non early abort? 
I think you should make this configurable via a flag on the bazel target or a flag to ibazel itself (maybe both)
The default should remain the same so the change is not breaking

mistic: For the time being, and at least to begin with, it can be achieved through the usage of a ENV VAR `IBAZEL_ABORT_COMPILATION_EARLY=1` 😃

tylersmalley:
Thanks everyone for your help so far on this change.

What is needed or what could we do to pull this over the finish line? We're nearing a point in our Bazel migration of Kibana that we will be blocked from not having this functionality.

mistic:
@achew22 can we move on with the review on this PR?

Wenqer:
@achew22 Is there any chance that this PR will be accepted?

PR #494, opened 2021-12-06
fix(NA): support bazel paths with whitespaces on windows

mistic:
Re-opening of https://github.com/bazelbuild/bazel-watcher/pull/480

This PR is a fix for a thing I found while testing #479 on Windows.

My usually setup is to install bazelisk within npm and then use it to manage bazel while ibazel is also installed through npm. On this setup bazelisk endup being installed on C:\Progam Files which contains a white space. That setup causes an error to be thrown while querying for files to be watched which doesn't allow to use ibazel. The changes in this PR fixes the problem.


mistic:
@achew22 can we move on with the review on this PR?

th0masb:
@mistic thanks for this fix, we've tested a build from this branch and it solves the issue we were running into. Would be good to get this merged.

bazel/bazel_windows.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// windows specific as it assures the bazelPath is always double quoted
+	// which assures we can support paths with whitespaces.
+	// It works by specifying the CmdLine after the exec command has been specified
+	// and by wrapping in double quotes the bazelPath content
+	//
+	// NOTE: SysProcAttr.CmdLine does not exist/is supported to be compiled on other
+	// OS other than Windows which is the reason why newCommand fn was created both
+	// for Windows and Unix
+	bazelPath = fmt.Sprintf("\"%s\"", bazelPath)
```
achew22, line 57: If this is all that's necessary to make it work in windows, make a function like `quote(s string) string { return fmt.Sprintf("%q", bazelPath) }` and call it `bazel.go`. There is no need to break out the entire `newCommand` just to change this one line.

achew22: Okay, then you can do

```
// in bazel_windows.go
func setProcessAttributes(cmd *exec.Command, bazelPath string, args []string) {
  b.cmd.SysProcAttr = &syscall.SysProcAttr{}
  b.cmd.SysProcAttr.CmdLine = fmt.Sprintf("%s %s", fmt.Sprintf("%q", bazelPath), strings.Join(args[:], " "))
}
```

```
// in bazel_unix.go
func setProcessAttributes(cmd *exec.Command, bazelPath string, args []string) {}
```

I'm very reticent to have large swaths of code duplicated for multiple OSes. This will compile down to be identical in byte code but will be much easier to grok the windows specific stuff.

mistic: Done

PR #531, opened 2022-07-27
Display output runner command in prompt, rather than output match

kmicklas:
It's a bit confusing that the prompt asks about executing a command, but what is printed is not the command, but the output match which triggered it. It looks like this may be from the buildozer use case?


achew22:
Looks good to me. Thanks!

PR #532, opened 2022-07-28
Support concatenating static text with substitutions in output runner

kmicklas:
Our use case: sometimes we want to run gazelle on a directory whose full path can't be extracted directly from the output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,11 @@ it detects a missing import for your go code.
     "regex": "^buildozer '(.*)'\\s+(.*)$",
     "command": "buildozer",
     "args": [ "$1", "$2" ]
+  },
+  {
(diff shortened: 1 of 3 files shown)


achew22 (changes requested):
The code + tests looks good to me! Could I have you add a couple of examples to the [readme](https://github.com/bazelbuild/bazel-watcher/blob/master/README.md#output-runner) so that you're not the only one who gets to enjoy this new feature?

Happy to merge at that point and do a release for you. Thanks!

achew22 (approved):

achew22:
This has also been released at https://github.com/bazelbuild/bazel-watcher/releases/tag/v0.18.0.

Thanks so much for your contribution!

PR #537, opened 2022-08-14
Updaing the bazel build instructions.

xuehuichao:
After the new structure, the old bazel build command stops working because the target no longer exists. Updating the instructions to reflect the new structure.

Tested:

  cd "$(mktemp -d)"
  curl -L
  https://github.com/bazelbuild/bazel-watcher/archive/master.zip -o
  master.zip
  unzip master.zip
  cd bazel-watcher-master
  bazelisk build //cmd/ibazel:ibazel
  ls -lh "$PWD/bazel-bin/cmd/ibazel/ibazel_/ibazel"

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,13 @@ You can, of course, build iBazel using Bazel.
 ```
 $ git clone git@github.com:bazelbuild/bazel-watcher
 $ cd bazel-watcher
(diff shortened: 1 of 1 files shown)


achew22:
Thanks for fixing this!

PR #539, opened 2022-08-24
Allow --nohome_rc as a startup flag to bazel

alex-torok:
`--nohome_rc` can be useful when you want to be sure that nothing in your home bazelrc file is interfering with a project. Add it to the list of overridable startup flags

see: https://bazel.build/reference/command-line-reference#flag--home_rc


achew22:
Thanks so much!

PR #547, opened 2022-09-17
Pass through necessary args for cquery

philsc:
`ibazel` runs an initial `cquery` on startup. Unfortunately, it does
not pass through command line arguments other than
`--override_repository`. That means we're not passing through things
like `--platforms`. That in turn means the target being queried can
actually be incompatible and the query will fail.

This patch fixes the situation by passing all bazel options through to
the `cquery`.

The new test validates just 2 of the ways a `cquery` requires more
command line arguments than what it has been getting.


achew22:
What a great contribution! It even has an end to end test. Much appreciated @philsc.

Issue #448, opened 2020-12-16
Support ibazel coverage


Toxicable:
Hey team, we're looking at integrating coverage into our repository with bazel and we're after continuous updates to our coverage.
Currently we need to develop our tests with `ibazel run :target` then separately we run `bazel coverage :target`, however ideally we'd just run `ibazel coverage :target` to get fast updates to our code coverage

achew22:
Oh how I wish there was a team behind this 🤣. Sorry, it's just me.

Alright, here is how I would go about doing this:

1) Add a new `Test` method into `ibazel.go` called `Coverage`

  https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/ibazel/ibazel.go#L264-L267

2) Add a new case statement to handle `coverage` being passed in on the command line:

  https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/ibazel/main.go#L198-L199

3) Add to the e2e helper a new method `Coverage`

  https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/e2e/ibazel.go#L58-L61

4) And the associated internal `coverage` function based on

  https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/e2e/ibazel.go#L266-L283

5) Add a new test case called `TestSimpleCoverage` shaped like this:

  https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/e2e/simple/simple_test.go#L65-L71

  Please note that you'll probably need to add an `sh_test` to the test fixture. The fixtures are formatted as [txtar](https://godoc.org/golang.org/x/vgo/vendor/cmd/go/internal/txtar#:~:text=Txtar%20format,the%20next%20file%20marker%20line).

 https://github.com/bazelbuild/bazel-watcher/blob/35ab9a9994dea4bc2822039c3617da28317094f0/e2e/simple/simple_test.go#L15-L57

It looks like I don't have any test coverage of `ibazel test` in there. If you were feeling like a total boss it would be amazing if you did the `e2e` stuff also for a new test.

PR #587, opened 2023-02-02
Add bazel coverage support

Lorite:
Fixes issue #448. I more or less followed the comments in the issue from @achew22. Then I added code to some other parts that were required. Let me know if I missed something else.
Apart from that, I also added a couple of very simple tests for the test and coverage commands to internal/e2e/simple/simple_test.go.

### DIFF
--- a/internal/e2e/ibazel.go
+++ b/internal/e2e/ibazel.go
@@ -74,6 +74,34 @@ func (i *IBazelTester) Test(bazelArgs []string, targets ...string) {
 	i.cmd = exec.Command(ibazelPath, args...)
 	i.t.Logf("ibazel invoked as: %s", strings.Join(i.cmd.Args, " "))
 
+	// print hello world
+	i.t.Logf("Hello World")
+
+	i.stdoutBuffer = &Buffer{}
+	i.cmd.Stdout = i.stdoutBuffer
+
+	i.stderrBuffer = &Buffer{}
+	i.cmd.Stderr = i.stderrBuffer
+
+	if err := i.cmd.Start(); err != nil {
+		i.t.Fatalf("Command: %s", i.cmd)
+	}
+}
+
+func (i *IBazelTester) Coverage(bazelArgs []string, targets ...string) {
+	i.t.Helper()
+
+	args := []string{"--bazel_path=" + i.bazelPath()}
+	args = append(args,
+		"--log_to_file="+i.ibazelLogFile,
+		"--graceful_termination_wait_duration=1s")
+	args = append(args, "coverage")
+	args = append(args, "--bazelrc=/dev/null")
+	args = append(args, targets...)
+	args = append(args, bazelArgs...)
+	i.cmd = exec.Command(ibazelPath, args...)
+	i.t.Logf("ibazel invoked as: %s", strings.Join(i.cmd.Args, " "))
+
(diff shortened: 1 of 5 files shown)


internal/e2e/simple/simple_test.go
```diff
@@ -54,6 +64,10 @@ sh_binary(
 )
 -- subdir/subdir.sh --
 printf "Hello subdir!"
+-- test_simple_test_failing.sh --
+echo "1"
+-- test_simple_test_passing.sh --
+echo "0"
```
achew22, line 70: Is there any way to have this test print out something to indicate that it is running (or not running) in coverage mode?

achew22 (changes requested):

internal/e2e/simple/simple_test.go
Lorite: Not sure. We would probably need a coverage report generator... Something like:
`bazel coverage --combined_report=lcov --coverage_report_generator=<an sh_library rule> <a cc_test rule>`
or maybe some other bazel targets to be able to add these arguments to `bazel coverage`:
`--coverage_output_generator (a build target label; default: "@bazel_tools//tools/test:lcov_merger")
  --coverage_report_generator (a build target label; default: "@bazel_tools//tools/test:coverage_report_generator")
  --coverage_support (a build target label; default: "@bazel_tools//tools/test:coverage_support")
  --[no]collect_code_coverage (a boolean; default: "false")`

Lorite: Let me know if you know a better approach.
I tested this implementation on a real project and it works perfectly.

internal/e2e/ibazel.go
```diff
@@ -74,6 +74,34 @@ func (i *IBazelTester) Test(bazelArgs []string, targets ...string) {
 	i.cmd = exec.Command(ibazelPath, args...)
 	i.t.Logf("ibazel invoked as: %s", strings.Join(i.cmd.Args, " "))
 
+	// print hello world
+	i.t.Logf("Hello World")
```
achew22, line 78: Do we need this?

internal/e2e/simple/simple_test.go
```diff
@@ -54,6 +64,12 @@ sh_binary(
 )
 -- subdir/subdir.sh --
 printf "Hello subdir!"
+-- test_simple_test_failing.sh --
+echo "I'm a failing test."
+exit 1
+-- test_simple_test_passing.sh --
+echo "I'm a passing test."
```
achew22, line 71: if you invoke `env` in here are there any environment variables that differ between a `test` run and a `coverage` run?

internal/e2e/ibazel.go
Lorite: My bad. I removed it.

internal/e2e/simple/simple_test.go
Lorite: I tried with printenv in both tests and all the environment variables were the same. I can't find a better way to test it than what I did.

Lorite:
Sorry for taking so long to fix the PR. I was busy with other stuff :/

internal/e2e/simple/simple_test.go
achew22: I think this is as good as it is going to get

achew22 (approved):

PR #593, opened 2023-03-10
feat: add detection for isatty

sxlijin:
This is subpar compared to what Bazel can autodetect, but it's better than nothing.

Bazel's `--curses=auto` is implemented by [passing to itself a flag `--isatty=`](https://github.com/bazelbuild/bazel/blob/bd011cdeddeb9750468612bb89ea9f189fcbd726/src/main/cpp/option_processor.cc#L661-L662), which is computed using [platform-specific logic](https://github.com/bazelbuild/bazel/blob/bd011cdeddeb9750468612bb89ea9f189fcbd726/src/main/cpp/blaze_util_posix.cc#L770-L786).

Upgrade build directives to `//go:build` since this directive is now preferred since go 1.18.

Borrows from [mattn/go-isatty](https://github.com/mattn/go-isatty).

### DIFF
--- a/cmd/ibazel/main.go
+++ b/cmd/ibazel/main.go
@@ -51,6 +51,7 @@ var overrideableBazelFlags []string = []string{
 	"--dynamic_mode=",
 	"--features=",
 	"--flaky_test_attempts=",
+	"--isatty=",
 	"--keep_going",
 	"-k",
 	"--nocache_test_results",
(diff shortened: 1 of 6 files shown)


cmd/ibazel/main.go
```diff
@@ -196,8 +197,24 @@ func main() {
 	handle(i, command, args)
 }
 
+func applyDefaultBazelArgs(bazelArgs []string) []string {
+	for _, bazelArg := range bazelArgs {
+		if strings.HasPrefix(bazelArg, "--isatty=") {
+			return bazelArgs
+		}
+	}
+	if (isTerminal()) {
```
achew22, line 206: What are the consequences of turning this on? What does `--isatty` do to bazel? How does it behave differently? I'm having a hard time finding docs that describe the different behavior

sxlijin: It's used to control the behavior of `--color=auto` and `--curses=auto`:

* [flag definition](https://github.com/bazelbuild/bazel/blob/49731660d8348d4ce243c4531bd5f3b14006f1df/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java#L173-L184)
* [flag references](https://github.com/bazelbuild/bazel/blob/49731660d8348d4ce243c4531bd5f3b14006f1df/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java#L284-L290)

achew22 (approved):

PR #595, opened 2023-03-28
fixes #594: Replace ibazel query with cquery

### LINKED ISSUE #594: ibazel ought to use CQuery instead of regular query
**Describe the bug**
Running `bazel build <target>` then `bazel query <target>` invalidates the cache sometimes because `bazel query` doesn't consider select statements.

**Reproduction instructions**
Any repo with a target with select statements. Run the above commands and see the cache is invalidated.

**Expected behavior**
The cache should not be invalidated. This leads to really long boot up times.

 - OS: `MacOS 13.3`
 - Browser: N/A
 - ibazel Version: `v0.16.2`
 - Bazel version: `6.1.0`

jakeleventhal:


achew22:
@jakeleventhal Thanks for the contribution! Lemme kick off a tag and release really quick

PR #598, opened 2023-03-29
Removed ibazel buildfiles query

jakeleventhal:


internal/e2e/local_repository/local_repository_test.go
```diff
@@ -86,22 +86,6 @@ func TestMain(m *testing.M) {
 	})
 }
 
-func TestRunWithModifiedFile(t *testing.T) {
```
achew22, line 89: Why are you removing this test?

achew22 (changes requested):
It seems there are still some outstanding issues in CI. Could you take a look at that?

internal/e2e/local_repository/local_repository_test.go
jakeleventhal: commented out to see if CI passed. doesnt seem to be the case. having trouble figuring out why this is failing. it appears to not be updating the file even though we are only changing the .sh file

jakeleventhal: Update: `TestRunWithModifiedFile` is the failing test in question. With the `i.watchFiles(fmt.Sprintf(buildQuery, joinedTargets), i.buildFileWatcher)` line present (as is on master), the test passes. When removed, that test fails.

I added a print statement to ibazel.go::watchFiles
Screenshot 2023-03-29 at 12 14 21 AM

Strangely, when the buildQuery line is present, the printed output is
```
we are watching /private/var/tmp/.../bazel_testing/bazel_go_test/main/test.sh via kind('source file', deps(set(//:test)))
we are watching /private/var/tmp/.../bazel_testing/bazel_go_test/secondary/lib.sh via kind('source file', deps(set(//:test)))
```
When the line is commented out, the log saying we are watching `lib.sh` is no longer present. You would think this has no effect because:
1. `buildfiles` shouldn't affect queries of .sh files
2. Since both log lines say `via kind('source file', deps(set(//:test)))`, you would think that the other line should have its query remain unaltered.

sluongng:
Small nits on this and the previous PR: it would be nice to have a bit more description in both PRs (or in the commit message).

A quick summary from the Slack conversation leading up to these changes would suffice.
It would mean a lot to folks who have to maintain an internal mirror of this repo/tool to know what/why a change was made.

Thanks!

jakeleventhal:
> Small nits on this and the previous PR: it would be nice to have a bit more description in both PRs (or in the commit message).
> 
> A quick summary from the Slack conversation leading up to these changes would suffice. It would mean a lot to folks who have to maintain an internal mirror of this repo/tool to know what/why a change was made.
> 
> Thanks!

Updated https://github.com/bazelbuild/bazel-watcher/pull/595 and this PR

internal/e2e/local_repository/local_repository_test.go
sluongng: I think there are 2 things to do here.

1. Long term solution: Check with bazel team over at bazelbuild/bazel to see why `buildfiles()` is not supported in cquery. Imo there is no obvious reason why such function is ok in query but error out in cquery (which should have more data).

2. Short termsolution : Consider detecting `buildfiles()` in `queryForSourceFiles` and switch to using query instead of cquery.

It makes a lot less sense that we would remove watching of buildfiles, considering that user could be modifying a starlark file and require rebuild of the target anyway.

PR #604, opened 2023-04-18
Fix buildfiles query

### LINKED ISSUE #603: BUILD file changes not detected anymore
**Describe the bug**
BUILD file changes are not detected by bazel-watcher and when bazel-watcher is started it non-fatally emits the following error:
```
ERROR: Error doing post analysis query: Evaluation failed: buildfiles() doesn't make sense for the configured target graph
```

The issue is very likely related to this change: https://github.com/bazelbuild/bazel-watcher/commit/ac2ce9f7406798b10c584efbac14ca05ba8cd818

The change moves the queries bazel-watcher does over to cquery but cquery does not support `buildfiles` in it's queries.

**Expected behavior**
BUILD file changes should be detected

 - OS: Fedora 37
 - ibazel Version v0.22.0
 - Bazel version 6.1.1

purkhusid:
## What?
Commit ac2ce9f7406798b10c584efbac14ca05ba8cd818 introduced a change where bazel-watcher uses cquery instead of query to get the files to watch. This works well for source files but the `buildfiles` keyword is not supported in cquery so the query for BUILD files fails.

This change uses cquery for source files and query for BUILD files.

Fixes #603

### DIFF
--- a/internal/ibazel/BUILD
+++ b/internal/ibazel/BUILD
@@ -36,6 +36,7 @@ go_library(
         "//internal/ibazel/profiler",
         "//internal/ibazel/workspace",
         "//third_party/bazel/master/src/main/protobuf/blaze_query",
+        "//third_party/bazel/master/src/main/protobuf/analysis",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


achew22:
While this does fix the issue, it still issues a `bazel query` which means that we'd restart the bazel server. Since this was the primary motivation for the query, I think a safer strategy is going to be to roll back the change in question and take another stab at it. Thanks for sending in your PR, it's hugely appreciated!

PR #621, opened 2023-08-01
feat(keep_going) Support keep going when querying for source files 

ankit-agarwal1999:
Ran into an issue with ibazel where if --keep_going was enabled, ibazel would not watch on the files that were ouputted by keep_going, it would just fail if the query was faulty. This is because the watch query didn't support the keep_going flag and didn't look at the exit code of the query. This PR adds that functionality to ibazel. Now querying source files will still work if keep_going is enabled. 

Looks like there aren't tests to Query and CQuery in the codebase right now, so let me know if you want me to add some. I'll keep the PR as is for now.


ankit-agarwal1999:
cc @achew22

ankit-agarwal1999:
/github-actions not stale just waiting for review :)

PR #623, opened 2023-08-17
RFC: Re-trigger build query on file-set changes in BUILD directories

gzm0:
Attempt to make #135 a bit less bad.


gzm0:
I wanted to write a proposal first, but it seemed the code was simple enough.

- Would you be open for such a change?
- What performance considerations do we need to make? / Do we need to flag guard this?

Note: A proper-er solution would of course watch the entire package, for this we'd either need to:
- Wait on https://github.com/fsnotify/fsnotify/issues/18
- Build recursive watching ourselves (would have the advantage that we could stop at subpackage boundaries).

PR #627, opened 2023-09-29
fix: use go:embed  instead of go_embed_data

moisesvega:
In the upcoming update, '[rules_go](https://github.com/bazelbuild/rules_go/releases)' will no longer include the 'go_embed_data' rule. Instead, this diff modifies 'go_embed_data' to adhere to the recommended approach for embedding data using 'go:embed'.


achew22 (approved):

achew22:
Thanks so much for fixing this!

PR #638, opened 2023-11-27
remove log statement for legacy bazel watcher on macOS

joeljeske:
This fsevents file watcher has been in use for macOS for a long time and can be made default without log spamming. Allow people to still easily drop to the legacy watcher if desired.


achew22 (approved):

achew22:
@joeljeske thanks for sending this in! Should clean up the experience for people

Issue #646, opened 2024-03-07
support `MODULE.bazel` for identifying workspace root


klandergren:
**Describe the bug**
`ibazel` does not appear to support finding the workspace root of a project with _just_ a `MODULE.bazel` file (e.g. no `WORKSPACE` or `WORKSPACE.bazel`). As far as I can tell this is a valid repo structure for bazel 7. 

**Reproduction instructions**
[klandergren/ibazel-no-workspace-file](https://github.com/klandergren/ibazel-no-workspace-file) shows a minimal go project that works using bazel 7 but does not work with `ibazel` Version 0.24.0.

1. `git clone https://github.com/klandergren/ibazel-no-workspace-file`
2. `ibazel run //hello:hello`
3. See error

```console
iBazel [6:27PM]: Querying for files to watch...
iBazel [6:27PM]: Error finding workspace: ibazel was not invoked from within a workspace

iBazel [6:27PM]: Error querying for source files: ibazel was not invoked from within a workspace

iBazel [6:27PM]: Error finding workspace: ibazel was not invoked from within a workspace

iBazel [6:27PM]: Error querying for source files: ibazel was not invoked from within a workspace

iBazel [6:27PM]: Running //hello:hello
Computing main repo mapping:
Loading:
Loading: 0 packages loaded
Analyzing: target //hello:hello (0 packages loaded, 0 targets configured)
INFO: Analyzed target //hello:hello (0 packages loaded, 671 targets configured).
 checking cached actions
INFO: Found 1 target...
 checking cached actions
 checking cached actions

INFO: Elapsed time: 0.373s, Critical Path: 0.00s
 checking cached actions
INFO: 0 processes.
 checking cached actions
INFO: Build completed successfully, 0 total actions
INFO: Analyzed target //hello:hello (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //hello:hello up-to-date:
  bazel-bin/hello/hello_/hello
INFO: Elapsed time: 0.532s, Critical Path: 0.29s
INFO: 6 processes: 4 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 6 total actions
iBazel [6:27PM]: Starting...
iBazel [6:27PM]: Error finding workspace: ibazel was not invoked from within a workspace

Kips-MBP:ibazel-no-workspace-file klandergren$ 2024/03/06 18:27:57 hello world
```

**Expected behavior**
`ibazel` to find the workspace root via `MODULE.bazel`

[...]

PR #647, opened 2024-03-07
support MODULE.bazel file

klandergren:
addresses #646 

 * verified that `ibazel` with this patch can correctly watch and run the target in the linked issue
 * verified that `//internal/ibazel/workspace:workspace_test` passes

### DIFF
--- a/internal/ibazel/workspace/workspace.go
+++ b/internal/ibazel/workspace/workspace.go
@@ -59,6 +59,10 @@ func (m *MainWorkspace) FindWorkspace() (string, error) {
 			return path, nil
 		}
 
+		if _, err := os.Stat(filepath.Join(path, "MODULE.bazel")); err == nil {
+			return path, nil
+		}
+
 		// If we've reached the root, then we know the cwd isn't within a workspace
 		if path == volume {
 			return "", errors.New("ibazel was not invoked from within a workspace\n")
--- a/internal/ibazel/workspace/workspace_test.go
+++ b/internal/ibazel/workspace/workspace_test.go
@@ -73,6 +73,14 @@ func TestAppleCaseInsensitivity(t *testing.T) {
 			files:         []string{},
 			err:           false,
 		},
+		"MODULE.bazel": {
+			startingWD:    "",
+			wantPath:      "",
+			dirs:          []string{},
+			workspacePath: "/MODULE.bazel",
+			files:         []string{},
+			err:           false,
+		},
 	}
 
 	startDir, err := os.Getwd()


achew22:
I think this is a great change! Would you be willing to add a new e2e test based on the [`simple`](https://github.com/bazelbuild/bazel-watcher/tree/master/internal/e2e/simple) test case but instead of having the test change `BUILD` files, having it change the `MODULE.bazel` file? I'd be very happy to merge as soon as something like that was in

klandergren:
Thanks @achew22! What specific changes to a `MODULE.bazel` should the new e2e test exercise?

I ask because I noticed that none of the tests in [simple_test.go](https://github.com/bazelbuild/bazel-watcher/blob/master/internal/e2e/simple/simple_test.go) test modifications to `WORKSPACE`, which is the closest equivalent file to what `MODULE.bazel` replaces. My understanding is that both `WORKSPACE` and `MODULE.bazel` are used to do things like load and resolve external dependencies which—I think—would involve external network requests which would not be desired for testing. 

I considered loading a macro via `load` but after getting an error found that only [a subset of methods](https://bazel.build/rules/lib/globals/module) are available for use in `MODULE.bazel` files.

In light of this my opinion is that an e2e test exercising `MODULE.bazel` may not be necessary but of course I defer to you.

klandergren:
I did some experimentation and investigation and found that [`sourceQuery`](https://github.com/bazelbuild/bazel-watcher/blob/be201b413f68f721408a4a615d57a568aa82642f/internal/ibazel/ibazel.go#L63) and [`buildQuery`](https://github.com/bazelbuild/bazel-watcher/blob/be201b413f68f721408a4a615d57a568aa82642f/internal/ibazel/ibazel.go#L64) are the two queries that control which files `ibazel` pays attention to when detecting changes. While it does pass the parent directories of these outputs to `fsevents` it [filters events (by path) against the results of these two queries](https://github.com/bazelbuild/bazel-watcher/blob/be201b413f68f721408a4a615d57a568aa82642f/internal/ibazel/ibazel.go#L286). Neither result contains `WORKSPACE` or `MODULE.bazel`, so to support an end-to-end test that detects changes in those files would be more involved and probably need its own separate PR.

farhaven:
@achew22 Is there a way to get this PR merged?

achew22 (approved):

n1ght-hunter:
@klandergren any chance you get the builds working so this pr can be merged?

farhaven:
@achew22 Is there a way to get more detailed logs from the failed Windows build? It looks like it just aborts at some point without a specific error message. Or maybe that test is just flaky and needs to be triggered again?

achew22:
The only additional information I have been able to find is on the artifacts tab. Unfortunately, windows remains hostile to people wanting to develop software for it.

alexeagle:
@achew22 could we re-open this? Bazel 8 was released today, and this issue is the last reason many projects will have a non-functional WORKSPACE file

PR #665, opened 2024-07-11
Add --remote_download_outputs to list of overridable bazel flags

gauntface:


achew22 (approved):

gauntface:
@achew22 thank you for the review - it looks like the mac builds are having issues, looking at the logs it's not immediately clear to me what the issue is, is it something resulting from this PR?

achew22:
I completely agree with your assessment that you didn't cause this. I'm struggling to come up with a reason why this would work in some platforms but not in others. 🤔 Even more so, why would it be `incompatible_by_default`. I'm cloning this onto my local box to see if I can reproduce and will report back

PR #672, opened 2024-10-24
use cquery for resolving files and dependencies

tobiaskohlbau:
This takes antoher stub at #594 by introducing cquery back in. In difference to the previous attempt this splits up the resolving of buildfiles. CQuery does not support buildfiles but without cquery bazel-watcher could watch files which are not reflected in the build for e.g. dependencies behind a select statement. In order to avoid depending on such cases first a cquery is issued which resolves dependencies of selected targets. These dependencies are then injected into the buildfiles query in order to get dependent build files.


tobiaskohlbau:
not stale

Issue #683, opened 2024-12-03
feat: support `.bzlmod` extension


austinwoon:
`bzlmod` extension will replace legacy `WORKSPACE` system as noted here: https://bazel.build/external/migration

the current `FindWorkspace` method does not find `WORKSPACE.bzlmod` file and will throw a false positive error of `ibazel was not invoked from within a workspace`

PR #697, opened 2025-01-09
feat: detect bzlmod files identifying workspace directory

chickenandpork:
In #683 `ibazel` cannot find a workspace by `WORKSPACE.bzlmod`
In my own work, `ibazel` cannot find the workspace by `MODULE.bazel`

Generally, the workaround has been to create an empty file `WORKSPACE.bazel` (which appeared with #536, shipped in v0.25.3) to help `ibazel` but this is likely not the expected behaviour.

Solution is proposed in this PR:
- looks for `WORKSPACE.bzlmod`, `WORKSPACE.bazel`, `MODULE.bazel`, falling back to `WORKSPACE`
- unittests for all four
- preserves the `IsDir()` check to protect against directories in path matching case-insensitive to these names
- adds the `WORKSPACE.bazel` (#536) to the same case-insensitive-aware check and treats consistently to the other files
- unittest for case-insensitive-matching directory: ignoring directory-matches in general

This solution is broader than proposed by @austinwoon (#684) -- but I only saw #684 after solving my own pain for myself.  I would prefer @austinwoon be recognized for their contribution; I will happily rebase on #684 if it helps establish austinwoon in the contribution path.

## Test
A simple test can be down with `rules_gazebo`, a small-ish ruleset that has workarounds:
0. assuming you're working in $HOME/src
1. `(cd ~/src && git clone git@github.com:bazelbuild/bazel-watcher.git)`
2. `(cd ~/src && git clone https://github.com/gazebosim/rules_gazebo)`
3. `(cd ~/src/bazel-watcher && bazel build //cmd/ibazel)`
4. `sudo install -m0755 ~/src/bazel-watcher/bazel-bin/cmd/ibazel/ibazel_/ibazel /usr/local/bin/` (assuming it's on your `$PATH`)
5. `(cd ~/src/rules_gazebo && ibazel build //...)` -- we expect the workarounds to allow this to work
6. `rm -f ~/src/rules_gazebo/WORKSPACE.b*` -- delete the workarounds
7. `(cd ~/src/rules_gazebo && ibazel build //...)` -- repaired `ibazel` still works here (due to `MODULE.bazel`)

### DIFF
--- a/internal/ibazel/workspace/workspace.go
+++ b/internal/ibazel/workspace/workspace.go
@@ -39,6 +39,7 @@ func (m *MainWorkspace) FindWorkspace() (string, error) {
 	}
 
 	volume := filepath.VolumeName(path)
+	sentinel_filenames := []string{"WORKSPACE.bzlmod", "WORKSPACE.bazel", "MODULE.bazel", "WORKSPACE"} // search order
 
 	for {
 		// filepath.Dir() includes a trailing separator if we're at the root
@@ -46,17 +47,16 @@ func (m *MainWorkspace) FindWorkspace() (string, error) {
 			path = volume
 		}
 
-		// Check if we're at the workspace path
-		if s, err := os.Stat(filepath.Join(path, "WORKSPACE")); err == nil {
-			if !s.IsDir() && s.Name() == "WORKSPACE" {
-				// In macOS directories called "workspace" will match because the file
-				// system isn't case sensitive.
-				return path, nil
-			}
-		}
+		for _, sentinel := range sentinel_filenames {
+			// Check if we're at the workspace path
+			if s, err := os.Stat(filepath.Join(path, sentinel)); err == nil {
+				// In macOS directories called "workspace" will match "WORKSPACE"
+				// because the file system isn't case sensitive
 
-		if _, err := os.Stat(filepath.Join(path, "WORKSPACE.bazel")); err == nil {
-			return path, nil
+				if !s.IsDir() && s.Name() == sentinel {
+					return path, nil
+				}
+			}
 		}
 
 		// If we've reached the root, then we know the cwd isn't within a workspace
--- a/internal/ibazel/workspace/workspace_test.go
+++ b/internal/ibazel/workspace/workspace_test.go
@@ -71,7 +71,7 @@ func TestAppleCaseInsensitivity(t *testing.T) {
 		},
 		"no workspace in workspace-named path": {
 			startingWD: "c/WORKSPACE",
-			wantPath:   "/a",
+			wantPath:   filepath.FromSlash("/a"),
 			dirs: []string{
 				"a/b",
 				"c/WORKSPACE",
@@ -97,7 +97,7 @@ func TestAppleCaseInsensitivity(t *testing.T) {
 			// the "WORKSPACE" dirname should not early-quit the search, allowing
 			// us to find /c/MODULE.bazel
 			startingWD: "c/d/WORKSPACE",
-			wantPath:   "/c",
+			wantPath:   filepath.FromSlash("/c"),
 			dirs: []string{
 				"a/b",
 				"c/d/WORKSPACE",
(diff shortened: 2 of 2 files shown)


achew22:
Hey @chickenandpork! Thanks so much for your contribution. Now all we have to do is get that CI passing and I'd be delighted to merge. Thanks again!

chickenandpork:
> Hey @chickenandpork! Thanks so much for your contribution. Now all we have to do is get that CI passing and I'd be delighted to merge. Thanks again!

Awesome @achew22 -- problem is, I don't have anything running Windows to repro, but I think this will fix based on the log.

achew22:
Yeah, I don't have a Windows machine either. Sorry I can't be more helpful than to point you at the CI

achew22:
Sorry, my fingers are fat.

> Hey @achew22 I see a bit of flake in the testing: things I didn't change passed before, failed now. The testing looks for certain text strings from bazel, but I don't think bazel promises to keep entropy away from text strings.
>
> Has anyone looked at ignoring the output text that sometimes shows up?

I could be mistaken but I believe the strings it is looking for are emitted by iBazel. Give that we're in the same repo I'm ish-comfortable depending on them. If you see things that are from bazel and wanna rework the test to not depend on that, I'm very supportive.

chickenandpork:
Hey @achew22 
> I could be mistaken but I believe the strings it is looking for are emitted by iBazel. Give that we're in the same repo I'm ish-comfortable depending on them. If you see things that are from bazel and wanna rework the test to not depend on that, I'm very supportive.

This is a good sentiment; I appreciate that.

Bazel-watcher testing is looking for strings from Bazel.  Exactly certain strings. No more, no less.

Unfortunately, bazel occasionally adds helpful INFO messages in some cases… inconsistently, in a way that makes “no more, no less” into “flaky unittest”.  …I *think* that’s what’s happening.

I don’t think bazel expected that their stdout/stderr is suddenly an API, with specific static content, but Bazel-watcher unittests like it is.

I have frustration because Bazel has evolved and Bazel-watcher is not actually the same repo: one of the things I like about a monorepo is that the tests show what changes break other components.  Ironic that in tools favoured for monorepos, we need to do a polyrepo pattern of catching up to the error from a while back.

having joined a new role, I have the time to fix the issue, but I worry on the unknown demands of fixing the unittest to permit the fix to merge and make ibazel usable again.  I can try, but the new role may take priority.

do you use ibazel daily?  If so , you likely have peppered the workaround in your repos.  Would you consider running with this change, and stripping out the workaround?

chickenandpork:
In re-running the test, there is a change in behaviour from `bazel`, which causes some tests to fail.

Manually running `bazel test //internal/e2e/error:error_test --action_env=HOME --cache_test_results=no` causes ~every second test to fail.

Running with settings for flaky tests up to 5 shows 2 of 5, 3 of 5 to be failing but eventually a test will run without the extra output, and it passes.

Expanding the default timeout in `internal/e2e/ibazel.go` seems to get a better passing rate.  I'm pushed that as de9fb40 on my side, or 166fc67 

@achew22, is that too cowardly, to merely increase the timeout for this time?

chickenandpork:
Meanwhile, to unblock my day-job, https://github.com/chickenandpork/bazel-watcher/releases/tag/v0.25.3.1 is a bogus release just to carry the results of this work to a pre-built.  I've only uploaded `darwin` because that's all I have (buildingwith the native toolchain on amd64 and arm64), and the cross-compile is not correct yet.

alibimahdi:
We managed to use the workaround (WORKSPACE) but really we need a real solution

chickenandpork:
Hey @alibimahdi 
> We managed to use the workaround (WORKSPACE) but really we need a real solution

Yeah, the unittests are really stopping me -- they inconsistently seem to fail on text that `bazel` doesn't always put out: because bazel sometimes outs out "helpful" `INFO:` messages, sometimes the tests see different text.

I have been a bit occupied with ${day_job} to move this forward and I don't think @achew22 will permit this without a whole lot of work on the unittests to unblock.

chickenandpork:
#710 should allow this PR to go green; let's rebase this on #710 710.

chickenandpork:
Rebased on #710

achew22 (approved):

PR #709, opened 2025-03-21
chore: upgrade bazel to 7.4.1

janwinkler1:
strongly inspired by #703 with three important distinctions:
- we upgrade bazel version to 7.4.1 but do not enable bzlmod (yet) , #712 will do that
- actively disabling three e2e tests on macos darwin CI build steps (override_repository_test, many_dirs_test, and symlink_dir_test)
- add an ANSI stripper because ANSI sequences were causing issues in ExpectNoError

IMO at this stage this might be the way to go since CI is causing issues and the project seems to be stuck in a catch22 (even though #710 fixed CI issues, some new ones emerged upgrading bazel)

thus i suggest a pragmatic approach.  

#714 and #715 address tasks to be done once we have upgraded the bazel version and have a passing CI

PR #703, opened 2025-02-23
chore: Update bazel to 8.1.0

mrmeku:
Addtionally, migrate WORKSPACE depedencies to bzlmod.

Co-authored-by: me@andrewzallen.com

PR #710, opened 2025-03-21
fix: repair/update e2e tests: unblock CI

chickenandpork:
When working on #697 I found that the CI for bazel-watcher has grown a bit stale: it doesn't pass on the default branch (master) anymore.

Digging deeper, I found and fixed a few classes of problems:
 - some timeouts (default 20s) are no longer sufficient for Bazel to come online and sort things out
 - some bazel side-effects have changed: printing strings from `.bzl` imports don't always happen anymore
    - I did look at changing direct dependencies to trigger a change, but it wasn't sufficient

I did find that a blanket default of ~50s did make those timing-dependent tests return to green, but the test run took much longer, so I've individually trimmed an overriding timeout instead.

This PR allows the unit tests to complete locally, and should similarly unblock CI on the formal pipeline.  This should unblock existing RenovateBot PRs such as #570, #690, #706, #708, and can unblock research into #709 and #703.


chickenandpork:
@achew22 looks like this CI is failing here, but works fine locally.

I really see a challenge in CI failures that cannot be repro'd -- currently angry at the BCR presubmit process for the same reason.  Many of the current failures are timing issues.

achew22:
> some bazel side-effects have changed: printing strings from `.bzl` imports don't always happen anymore

Out of curiosity, is that because of something that changed in Bazel or is this some other kind of regression?

achew22:
When it comes to extending the timeouts to make CI pass, I think that's a very reasonable course of action. 

Apologies for how frustrating this is. I've not worked using Bazel for many years now, and not worked for G for many more.

I very much appreciate your efforts. I know this project has high SEO and is frequently linked. I'm glad it's useful to so many, and I know all those people appreciate the work you're putting into this.

chickenandpork:
> > some bazel side-effects have changed: printing strings from `.bzl` imports don't always happen anymore
> 
> Out of curiosity, is that because of something that changed in Bazel or is this some other kind of regression?

I feel like it's a change in what Bazel displays, and we were relying on behaviour that wasn't designed, but observed..  Perhaps the content is parsed (and executed during) only on demand where perhaps it was more liberally before?   A case of being better about build-avoidance eliminating the signal we were using :-/

chickenandpork:
@achew22 I've found that the build pipeline uses a command line option that doesn't exist.  I tried a few ways (such as 36bab9f) to mitigate.  How can I edit the build kite definition?

achew22:
> @achew22 I've found that the build pipeline uses a command line option that doesn't exist for query. I tried a few ways (such as https://github.com/bazelbuild/bazel-watcher/commit/36bab9f024415423ca3b2aeb4f5d5a6c58b757b9) to mitigate. How can I edit the buildkite definition?

I haven't worked with the CI stuff in more than half a decade, but I think it is defined in/around [here](https://github.com/bazelbuild/continuous-integration/blob/355e4f58d2c190548c1eb55f038f200c14899906/buildkite/terraform/bazel/main.tf#L1841)

chickenandpork:
@achew22 
> I haven't worked with the CI stuff in more than half a decade, but I think it is defined in/around [here](https://github.com/bazelbuild/continuous-integration/blob/355e4f58d2c190548c1eb55f038f200c14899906/buildkite/terraform/bazel/main.tf#L1841)

Yeah, that's bringing in [04ac42974](https://github.com/bazelbuild/continuous-integration/commit/04ac42974140aa348dc281939a45afd7ac4eec2e) which is putting the `jvmopt` in the outer call to `bazel`, which of course builds and tests `ibazel`, then `ibazel` runs `bazel` which is reading `/etc/bazel.bazelrc`

I think 04ac42974 line 1967 is a bit wrong: it's hitting `common_build_flags`, so `query` as well as `build`, but I don't think it directly affects us in this error.

I think I'm looking for where the macOS build hosts are configured.  Alternatively, I could try to delete the file, but I don't want to alter the build systems, and likely don't have permission from inside a pipeline.  Additionally, `nosystem_rc` is not available there.

Do you remember anyone I can talk to there?

I've raised this in slack as well:
https://bazelbuild.slack.com/archives/CA31HN1T3/p1743228317866509

chickenandpork:
@achew22 name-dropped you in https://github.com/bazelbuild/continuous-integration/issues/2201 posting about this issue.

I can't find where the actual `/etc/bazel.bazelrc` is defined.

chickenandpork:
@achew22 CI showing green; some immediate cleaning, but I hope we can merge this soon-ish to unblock other work.  I still intend (after) to trim back the timeouts and such for faster CI.  There's some other "catch-up" work to do as well.

I think we can piecemeal shift unit tests to bzlmod after this, or even test both.

ashwin153:
@chickenandpork Thanks for doing all this work. Can’t wait to use it!

achew22 (approved):

achew22:
Completely agree that we can work on smaller chunks. Having CI green again is a big step in the right direction though. Thank you for all your hard work!

chickenandpork:
Thanks @achew22 :)

PR #721, opened 2025-04-09
Download outputs when running in notify mode

DavidANeil:
If `--remote_download_outputs=minimal` is set, there is currently no good way to use an `ibazel_notify_changes` `run` target.

Alternative I considered: set an environment variable when invoking `bazel` from `ibazel`, and allow consumers to use a bazel wrapper to only set `--remote_download_outputs=minimal` when not being run from ibazel.  The committed solution seems better than this because it requires less setup from the end user.

This is a breaking change, because if a target previously relied on `--remote_download_outputs=all`, then this will break that expectation.

I'm willing to be told that this change is a bad idea, and I will just add it to our fork, but I thought I would send it here too.


DavidANeil:
Found a better way, new PR coming soon.

PR #712, opened 2025-03-23
chore: switch to bzlmod

janwinkler1:
- enable bzlmod
- adapting e2e tests (copy pasted from #703)
- conservatively increase timeouts for tests failing in macos CI step
- fixing race condition causing macos CI step to fail
- changes in go.mod were only necessary to get CI ubuntu steps passing, so we might want to upgrade CI ubuntu build steps soonish, tracked in issue #714

PR #723, opened 2025-04-09
chore: Re-enable macOS tests

chickenandpork:
This PR is intended to depend on @janwinkler1 's PR709 and PR712 and intends to exercise the CI process for changes to allow macos CI to be un-skipped.  This PR intends to exercise the CI so that I can repro the failures we're chasing.


chickenandpork:
@janwinkler1 has raised in [#712](https://github.com/bazelbuild/bazel-watcher/pull/712#issuecomment-2790518918) that the fix I offered as https://github.com/janwinkler1/bazel-watcher/pull/1 changing how `zlib` is built isn't necessary in this PR.

Absolutely correct.

We may not need it, either due to changes in dependencies due to shifting to bzlmod, or a newer version.  I *am* building on a `darwin/x86_64` so might not have visibility to that locally.  The OS toolchain is non-hermetic taint, so can change the implicit dependencies, giving a different result.  We're building with CI, so the toolchain may differ.  I think I should still build-check on `darwin/arm64` to ensure a home build on the latest Xcode tools still works.

I'm willing to do that as fast-follow to avoid gating @janwinkler1 on this double-check so long ad CI goes green.

chickenandpork:
Many of this PR is just changing timeouts.  I have been conservative in expanding timeouts as each linearly expands the total run time, but that's the complexity of fiddling with this right now.

I think the E2E test needs to be able to look for the desired strings before the timeout has expired so that it can early bail-out when success is reached.  This decouples the minimum runtime of each test form the timeout, allowing us to expand the timeout without always expanding the runtime of the test.  The timeout would become a back-to Prather than a guaranteed delay.

PR #740, opened 2025-05-16
Support overridden repositories.

ksnox:
The repository lookup logic no longer fails to find paths that have been altered by using overrides such as `local_path_override` or `git_override`.

### DIFF
--- a/internal/e2e/local_path_override/local_path_override_test.go
+++ b/internal/e2e/local_path_override/local_path_override_test.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -112,10 +111,6 @@ func TestMain(m *testing.M) {
 }
 
 func TestRunWithOverrideRepository(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skipf("Skipping windows tests.")
-	}
-
 	ibazel := e2e.SetUp(t)
 	ibazel.Run([]string{"--enable_bzlmod=1"}, "//:test")
 	defer ibazel.Kill()
--- a/internal/e2e/local_path_override/local_path_override_test.go
+++ b/internal/e2e/local_path_override/local_path_override_test.go
@@ -113,7 +113,7 @@ func TestRunWithOverrideRepository(t *testing.T) {
 
 	ibazel.ExpectOutput("hello!", 35 * time.Second)
 
-	if err := ioutil.WriteFile(filepath.Join(secondaryWd, "lib.sh"), []byte(secondaryRecompiledLib), 0777); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(secondaryWd, "lib.sh"), []byte(secondaryLib), 0777); err != nil {
 		log.Fatalf("Failed to write file lib.sh a second time. (%v)", err)
 	}
 	ibazel.ExpectOutput("hello!", 35 * time.Second)
--- a/internal/e2e/local_path_override/local_path_override_test.go
+++ b/internal/e2e/local_path_override/local_path_override_test.go
@@ -79,6 +79,7 @@ func TestMain(m *testing.M) {
 				"BUILD.bazel": secondaryBuild,
 				"lib.sh":      secondaryLib,
 				"MODULE.bazel":   secondaryModule,
+				"WORKSPACE":   "",
 			} {
 				if err := ioutil.WriteFile(filepath.Join(secondaryWd, file), []byte(contents), 0777); err != nil {
 					log.Fatalf("Failed to write file %q: %v", file, err)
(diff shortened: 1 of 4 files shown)


internal/ibazel/ibazel.go
```diff
@@ -603,3 +603,13 @@ func keys(m map[string]struct{}) []string {
 	}
 	return keys
 }
+
+func findInLocalRepository(localRepositories map[string]string, repo string) (string, bool) {
```
achew22, line 607: This is very interesting. I've not kept informed about what the current state of Bazel is on overrides. Two things I would like to see before I will merge this. 

1. Please add some comments that link to documentation of this behavior in the override context
2. I would also like to see some testing of this functionality to ensure it behaves in the right way

achew22 (changes requested):

internal/ibazel/ibazel.go
achew22: There are [a whole bunch](https://github.com/bazelbuild/bazel-watcher/tree/master/internal/e2e) of e2e tests that you could model your test off of as well

achew22: Looking at [The Bazel Query Reference](https://bazel.build/query/language) I'm not sure this gives me confidence on the translation scheme from `+` to actual filenames. Sorry, this is just a part of Bazel I have no familiarity with. Is there anything that might provide more context?

internal/e2e/local_path_override/local_path_override_test.go
```diff
@@ -0,0 +1,123 @@
...
+	deps = [
+		"@secondary//:lib",
+	],
+)
+-- test.sh --
+#!/bin/bash
+# Due to a lack of runfiles file support, search any potential override path.
+for f in '../secondary*/lib.sh'; do source $f; done
+# Including Windows.
+for f in '..\secondary*\lib.sh'; do source $f; done
```
achew22, line 49: Trying to be as helpful as I can here, but I know very little about windows development. First let's grab the germain part of the log.

```
C:\\b\\gx4a3jy3\\execroot\\__main__\\_tmp\\720a5f703af25788c71c4e9c9c043e06\\_bazel_b\\mj7aq3mf\\execroot\\_main\\bazel-out\\x64_windows-fastbuild\\bin\\test: line 3: ../secondary*/lib.sh: No such file or directory
     C:\\b\\gx4a3jy3\\execroot\\__main__\\_tmp\\720a5f703af25788c71c4e9c9c043e06\\_bazel_b\\mj7aq3mf\\execroot\\_main\\bazel-out\\x64_windows-fastbuild\\bin\\test: line 5: ..\secondary*\lib.sh: No such file or directory
       C:\\b\\gx4a3jy3\\execroot\\__main__\\_tmp\\720a5f703af25788c71c4e9c9c043e06\\_bazel_b\\mj7aq3mf\\execroot\\_main\\bazel-out\\x64_windows-fastbuild\\bin\\test: line 6: say_hello: command not found
```

https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/0196df81-453c-4400-9e24-c74153970d43/internal%5Ce2e%5Clocal_path_override%5Clocal_path_override_test%5Ctest_attempts%5Cattempt_1.log

achew22: Reading through [the docs](https://bazel.build/configure/windows#using-bazel-without-bash), I wonder if the buildkite runners have mingw or related installed

achew22: What if instead of running a bash script which seems to be a bit difficult here for some reason, we did something like:

```
genrule(
    name = "test",
    srcs = ["@secondary//:file"],
    outs = ["test.txt"],
    cmd = "cp $(location @secondary//:file) $@",
)
```

then you could `ibazel build` the file in the test, confirm the original value, overwrite the file and read the resulting output to confirm the new value. Since this is basically the "my first `genrule`" tutorial which I've seen a couple dozen times I'm pretty confident it'll work by whatever interpreter is failing on your slightly more complex (but still so simple it should work) script.

I don't believe that there is a commensurate operation to `ExpectOutput` for reading an output file, but I think something in `/internal/e2e/ibazel.go` like this may be useful.

```go
func (i *IBazelTester) ExpectOutputFile(path string, matcher *regexp.Regexp, delay ...time.Duration) {
	i.t.Helper()

	d := defaultDelay
	if len(delay) == 1 {
		d = delay[0]
	}

  var lastErr error
	stopAt := time.Now().Add(d)
	for time.Now().Before(stopAt) {
		time.Sleep(5 * time.Millisecond)

		b, err := os.ReadFile(path)
    if err != nil {
      lastErr = fmt.Errorf("os.ReadFile(%q): %v", path, err)
      continue
    }
    if matcher.Match(b) {
      return
    }
    lastErr = fmt.Errorf("ExpectOutputFile(%q): %v did not match the file. contents:\n\n%s", path, matcher, string(b))
	}

  if lastErr != nil {
    i.t.Error(err)
    return
  }
}
```

PR #744, opened 2025-06-11
Support overriding  --test_sharding_strategy

Aghassi:
In the event you want to watch a test target that is sharded, allow the end user to disable sharding so the target can be watched


achew22 (approved):

achew22:
Here go! https://github.com/bazelbuild/bazel-watcher/releases/tag/v0.26.2

PR #775, opened 2025-07-21
Fix notify_command.go to pass `args`allowing `--test_sharding_strategy=disabled` with `--`

Aghassi:
Fixes a bug introduced by #744 where on bazel run with sharding enabled the process would fail to launch. this also requires the user to use the `--` attribute before passing the flag so it gets sorted into args.

If this is not passed, the CLI will not discern between bazelArgs and args that come after `--`. Once this is landed, you can test a target with sharding so long as you do `ibazel run //path/to/target:test --` and then whatever flags you want.

### DIFF
--- a/internal/ibazel/command/notify_command.go
+++ b/internal/ibazel/command/notify_command.go
@@ -66,6 +66,8 @@ func (c *notifyCommand) Start() (*bytes.Buffer, error) {
 	b := bazelNew()
 	b.SetStartupArgs(c.startupArgs)
 	b.SetArguments(c.bazelArgs)
+	// For commands like `--test_sharding_strategy=disabled` which come through normal args
+	b.SetArguments(c.args)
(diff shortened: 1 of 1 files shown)


achew22:
Hey @Aghassi,

I'm interested in this change but unsure of how it affects things. Would it be possible to add a test case that demonstrates the old, improper behavior? I'd love to learn what's gone wrong here

Aghassi:
Sure I can try and add a test case. The basic gist from the logic I followed is that in this case this arg needs to be passed to `bazel run` and something about how the separation logic cascades basically filters this out of the run command. I presume because it looks like a test flag

Aghassi:
Not stale, I've just not had a chance to prioritize this relative to other work. This is still and issue that needs fixing. I need to provide a test and have not had a chance

PR #777, opened 2025-07-25
Fix building release binary for darwin_amd64

avdv:
Running the release job on macos-latest is using darwin_arm64, so cross compiling the
binary to darwin_amd64.

To make this work, either the build needs to run under Rosetta (effectively not cross compiling), or the build needs a
C cross compiler targetting darwin_amd64 (otherwise CGO is disabled).

Using the macos-13 runner fixes this problem by avoiding to cross compile.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,7 +35,7 @@ jobs:
 
           - name: darwin_amd64
             artifact: ibazel_darwin_amd64
-            os: macos-latest
+            os: macos-13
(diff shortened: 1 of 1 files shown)


achew22 (approved):

achew22:
@avdv, thanks for sending this in. I don't think the change as you have it hurts anything but it doesn't [seem to have resolved this issue](https://github.com/bazelbuild/bazel-watcher/actions/runs/16542583012/job/46785828774). I very much appreciate you trying!

PR #780, opened 2025-07-28
Fix CI release build for darwin x86_64

avdv:
Remove the `--cpu` flag which was added as a workaround ~3 years ago.

Note, I tested the build here: https://github.com/avdv/bazel-watcher/actions/runs/16563383352/job/46837547101


achew22 (approved):

achew22:
Looks like it was [successful](https://github.com/bazelbuild/bazel-watcher/actions/runs/16615462008). Thank you so much for your continued engagement fighting the entropy monster! Now, if only NPM could be less terrible

PR #816, opened 2026-02-08
Fix notify_command.go to pass argsallowing --test_sharding_strategy=disabled with --

Aghassi:
https://github.com/bazelbuild/bazel-watcher/pull/775

Reopens the above commit

```
Fixes a bug introduced by https://github.com/bazelbuild/bazel-watcher/pull/744 where on bazel run with sharding enabled the process would fail to launch. this also requires the user to use the -- attribute before passing the flag so it gets sorted into args.

If this is not passed, the CLI will not discern between bazelArgs and args that come after --. Once this is landed, you can test a target with sharding so long as you do ibazel run //path/to/target:test -- and then whatever flags you want.
```


achew22:
Could I have you (or Claude) add an e2e test to demonstrate that this failed before and now passes in the circumstance you're imagining?

PR #818, opened 2026-02-10
Use mapped repository names for watched files.

sbarfurth:
Bazel repositories have a canonical and a mapped/apparent name. The apparent name appears in labels whereas the canonical name is used in the file tree. The old logic was not taking mappings into account. This would lead to files being discard from watching due to their labels having the apparent name.

Example:

```starlark
bazel_dep(name = "my_cool_dep", repo = "foo")
```

This would lead to files inside the repository having labels starting with `@foo//...`. The canonical repo name is `my_cool_dep+` so the path on disk is something like `.../external/my_cool_dep+`. The old logic would look at the `bazel query` results and check for `my_cool_dep+` as the repository name, but it was actually `foo` which maps to `my_cool_dep+`. All files were discarded from watching and `ibazel` printed *Didn't find any files to watch from query*.


achew22:
Hrm, seems like probably a reasonable thing do to. Do you think you (or claude?) could add an e2e test to verify this functionality?

sbarfurth:
@achew22 added a test targeted at this specifically

sbarfurth:
*gentle ping* @achew22

achew22 (approved):

achew22:
Sorry about the delay. I'm a few thousand emails behind in my inbox these days, and I appreciate your patience

PR #822, opened 2026-03-18
use cquery for resolving files and dependencies

sbarfurth:
Use `cquery` for most query applications except for `BUILD` files. This makes queries more accurate to real builds since `select` statements and flags are evaluated. This should also prevent unnecessary cache invalidation from regular queries without flags.

Revival of #672. I've changed the original PR by fixing tests and quoting targets in the query file for better robustness. I manually tested this on our big repo which currently errors with `ibazel` due to `bazel query` traversing into places that would not ordinarily be on the action graph in builds.

### DIFF
--- a/internal/bazel/testing/mock.go
+++ b/internal/bazel/testing/mock.go
@@ -16,7 +16,10 @@ package testing
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
+	"io"
+	"os"
 	"os/exec"
 	"regexp"
 	"testing"
@@ -32,6 +35,7 @@ type MockBazel struct {
 	cqueryResponse map[string]*analysis.CqueryResult
 	args           []string
 	startupArgs    []string
+	info           map[string]string
 
 	buildError error
 	waitError  error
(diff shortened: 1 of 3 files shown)


achew22:
I have to admit that I have some FUD around this change. I believe cquery to be the correct thing to do but there have been some [attempts](https://github.com/bazelbuild/bazel-watcher/pull/595) that were [reverted](https://github.com/bazelbuild/bazel-watcher/pull/605) in the past. How can we ensure this is 100% definitely the way to go?

PS: I quite like the writing the query out to file so that there is no risk of it getting too long, that's a fun trick

sbarfurth:
> I have to admit that I have some FUD around this change. I believe cquery to be the correct thing to do but there have been some [attempts](https://github.com/bazelbuild/bazel-watcher/pull/595) that were [reverted](https://github.com/bazelbuild/bazel-watcher/pull/605) in the past. How can we ensure this is 100% definitely the way to go?
> 
> PS: I quite like the writing the query out to file so that there is no risk of it getting too long, that's a fun trick

I sympathize. I'd say the particular issue vector of the previous attempt is solved in this PR since it doesn't try to query `BUILD` files with `cquery`.

Additionally, this one passes all e2e tests. It seems to me that they're comprehensive enough to at least prevent an obvious breakage.

I also ran my code against our company monorepo where `ibazel` is currently unusable due to the usage of `bazel query` trying to analyze targets that cannot be analyzed. It worked fine. With these changes was the first time I was ever able to run `ibazel` on our repo without errors (though I also needed #818, but that's released).

achew22 (approved):
I guess the only real way to find out is to try it. Fingers crossed and thank you for your contribution!

PR #823, opened 2026-04-16
Fix labelsToWatch to keep workspace files after local-repo labels

neyo8826:
this fixes my js_run_devserver from aspect_rules_js

### DIFF
--- a/internal/e2e/local_repository/local_repository_test.go
+++ b/internal/e2e/local_repository/local_repository_test.go
@@ -33,19 +33,56 @@ function say_hello {
 }
 `
 
+const mainWorkspaceTailDep = `
+function say_local_tail {
+	printf "local-tail-1!"
+}
+`
+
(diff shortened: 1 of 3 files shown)


achew22:
Hrm... this might well be a necessary change, but I'm not 100% sure. Can you help me understand what isn't working and why you think this will resolve the issue?

neyo8826:
bazel query deps(...) outputs the list in local->remote order (// -> @...//) but cquery interleaves them, so the loop cannot stop there. it misses a lot of deps that way

achew22:
Okay, that seems reasonable. Can you add an e2e test that demonstrates the failure that is resolved by this change?

neyo8826:
added

neyo8826:
ping, still needed :)

longlho:
this is already covered in https://github.com/bazelbuild/bazel-watcher/pull/828 I believe, which is merged and released as 0.30.0

PR #826, opened 2026-04-22
fix: labelsToWatch short-circuiting faster than it should

### LINKED ISSUE #825: ibazel failing to restart JS application with heavy external dependencies
**Describe the bug**

Hey folks! I'm evaluating the use of `ibazel` with our monorepo and stumbled onto a relatively nasty bug where ibazel would fail to restart my relatively large JS application.

**Reproduction instructions**

https://github.com/markovejnovic/ibazel-js-repro

```bash
git clone https://github.com/markovejnovic/ibazel-js-repro.git
cd ibazel-js-repro
ibazel run //:dev
# wait until it prints 'hello world'
# edit src/main.ts to print a different string
# note ibazel doesn't catch it
```

**Expected behavior**

Ibazel restarts my target after a source file changes.

**Actual behavior**

Ibazel doesn't restart my target.

 - OS: macOS
 - ibazel Version: iBazel - Version 0.29.0, present on `HEAD`
 - Bazel version: 9.1.0
 
**Additional context**

I think the bug is in `ibazel.go:650`. When `labelsToWatch` filters for labels that should be watched, it has special handling for `localRepositories`. `aspect`'s external targets start with `@` and are referenced as `localRepositories`, which means that we hit the `break` in `ibazel.go:650`.

To be 100% honest, I don't really understand the intent behind this function and this was discovered by getting claude to sprinkle debug statements across the program, so I might be wrong in my guess, but changing the `break` to a `continue` does populate the `toWatch` list as I would expect.

markovejnovic:
Resolves #825. Please read that ticket as I don't really understand the intent behind this function nor its role in the larger application.

### DIFF
--- a/internal/ibazel/ibazel.go
+++ b/internal/ibazel/ibazel.go
@@ -636,7 +636,7 @@ func resolveLabels(labels []string, localRepositories map[string]string, workspa
 			if realPath, ok := localRepositories[repo]; ok {
 				label = strings.Replace(target, ":", string(filepath.Separator), 1)
 				toWatch = append(toWatch, filepath.Join(realPath, label))
-				break
+				continue
 			}
 			continue
 		}
--- a/internal/ibazel/ibazel.go
+++ b/internal/ibazel/ibazel.go
@@ -628,18 +628,7 @@ func (i *IBazel) watchFiles(toWatch []string, watcher common.Watcher) {
 	i.filesWatched[watcher] = filesWatched
 }
 
-func (i *IBazel) labelsToWatch(labels []string) ([]string, error) {
-	localRepositories, err := i.realLocalRepositoryPaths()
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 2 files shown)


achew22:
Hey @markovejnovic. Love the contribution. Can I have you put together an e2e test that can be included in the repo which demonstrates this behavior so that we don't regress going forward?

markovejnovic:
Of course! I'll try to get to it soon! Sorry I didn't reply earlier!

achew22:
I'd very much like to see an e2e test attached to this. Unit tests don't really demonstrate the issue since it's so easy to accidentally encode an incorrect assumption about the world

longlho:
this is already covered in https://github.com/bazelbuild/bazel-watcher/pull/828 I believe, which is merged and released as 0.30.0

markovejnovic:
Yep, should be fixed in #828, sorry all for dropping this. I got busy.

PR #827, opened 2026-06-01
Show when `bazel run` exited with code

KoltesDigital:
Hi,

As I'm using ibazel to develop and run a short-lived binary, I though it'd be good to have ibazel show when the binary has exited. A message "Starting..." is displayed at the beginning, this PR adds a message "Exited (code)" at the end.

I'm a bit frustrated that the OS can't tell exactly which child has exited, but AFAIK when there's a `Command` no other `Cmd` is spawned.

I have no idea how to write tests for that... and I hope it won't catch grandchildren.

### DIFF
--- a/internal/bazel/bazel.go
+++ b/internal/bazel/bazel.go
@@ -142,7 +142,6 @@ type Bazel interface {
 	Norun(args ...string) (*bytes.Buffer, error)
 	Test(args ...string) (*bytes.Buffer, error)
 	Run(args ...string) (*exec.Cmd, *bytes.Buffer, error)
-	Wait() error
 	Cancel()
 }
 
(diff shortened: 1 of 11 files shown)


KoltesDigital:
Well, the CI has shown the tests, and that there's no SIGCHLD on Windows. I guess the next solution is to spawn a thread that waits on the `Cmd`, and that would revert the signal handling.

KoltesDigital:
Ok now a goroutine is responsible for waiting, so it can't be done elsewhere: command structs gain two sync vars to report back, one atomic bool for whether it's running, and a channel to deal with the terminate timeout.

Bazel struct loses its Wait as it wasn't called anywhere, and I believe it's not exported into a library.

achew22:
Looks like a good start. Could you add some end to end tests to cover this new behavior so we don't accidentally regress it?

KoltesDigital:
@achew22 You're right, here's the e2e test.

However there's a data race I cannot get rid of. My guess it that the waiting goroutine from another test is still running when a new test starts. I've tried to ensure that the goroutine won't cross test boundaries (eg. calling `cmd.Terminate()` in `IBazel.Cleanup`, given that `terminate` should wait for it in any case) with no success. Do you have an idea about that?

PR #828, opened 2026-06-02
Fix shortened label list when querying files

KoltesDigital:
While making the other PR I stumbled across this bug by chance.

### DIFF
--- a/internal/ibazel/ibazel.go
+++ b/internal/ibazel/ibazel.go
@@ -647,7 +647,6 @@ func (i *IBazel) labelsToWatch(labels []string) ([]string, error) {
 			if realPath, ok := localRepositories[repo]; ok {
 				label = strings.Replace(target, ":", string(filepath.Separator), 1)
 				toWatch = append(toWatch, filepath.Join(realPath, label))
-				break
 			}
 			continue
 		}
--- a/internal/ibazel/ibazel_test.go
+++ b/internal/ibazel/ibazel_test.go
@@ -653,3 +653,47 @@ func TestParseTarget(t *testing.T) {
 		})
 	}
 }
+
+func TestIBazelConvertLabelsToPaths(t *testing.T) {
+	log.SetTesting(t)
+
+	i, mockBazel := newIBazel(t)
+	defer i.Cleanup()
+
+	outputBase := t.TempDir()
+	if err := os.Mkdir(filepath.Join(outputBase, "external"), 0o700); err != nil {
+		t.Errorf("failed to create external directory: %v", err)
+		t.FailNow()
+	}
+	mockBazel.SetInfo(map[string]string{
+		"output_base":  outputBase,
+		"install_base": t.TempDir(),
+	})
+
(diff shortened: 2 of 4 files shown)


internal/ibazel/ibazel.go
```diff
@@ -647,7 +647,6 @@ func (i *IBazel) labelsToWatch(labels []string) ([]string, error) {
 			if realPath, ok := localRepositories[repo]; ok {
 				label = strings.Replace(target, ":", string(filepath.Separator), 1)
 				toWatch = append(toWatch, filepath.Join(realPath, label))
-				break
```
achew22, line 650: This seems like a good change. Can you help me understand what exactly is being fixed here by adding an e2e test that demonstrates the underlying issue

KoltesDigital:
@achew22 I've added a unit test to directly test the offending `labelsToWatch` function. Since the break exits the loop for an external label, I've interleaved internal and external labels.
Edit: Windows doesn't follow symlinks.

internal/ibazel/ibazel_test.go
```diff
@@ -653,3 +653,47 @@ func TestParseTarget(t *testing.T) {
 		})
 	}
 }
+
+func TestIBazelConvertLabelsToPaths(t *testing.T) {
```
achew22, line 657: This is good, but it doesn't really confirm the true behavior of the failure. It confirms that it works the way the mock set it up. Is three any way to add an e2e test case to this?

KoltesDigital: I disagree. The wrong behavior is on the map-filter of a list. The mock acts as the filter condition. But the issue is that the map-filter implementation is broken. Bring back the break and you'll see the test fail. 

Regarding e2e I believe it'd be impossible to test because this behavior is not accurately exposed. I stumbled on the bug by chance, I noticed updating a particular file in a particular workspace triggered no rebuild, and I found the bug: it happened only because bazel query listed the file after other external files. But AFAIK the order of bazel query is not specified and might even change between invocations. From the CLI, there's no way to display query results and watched files, so no way to reliably detect that.

achew22: Sine that is the case then the periodic execution of the e2e test would demonstrate that it is flaky and result in a test failure, wouldn't it?

KoltesDigital:
@achew22 I've added a e2e test. I couldn't use `sh_library` like in local_repository or nested_module because Bazel 6.5.0 lists local files before external files... But then, I'm afraid the test only runs against 7.6.2, while current version is 9.1.1.

longlho:
hi guys any movement on this? we're encountering this bug as well

achew22 (approved):

achew22:
@KoltesDigital, I think that's okay. Thank you for the e2e test! Lemme kick off a release and get this out the door

@longlho thanks for commenting, this got lost in my email inbox and I wouldn't have surfaced it without your poke

longlho:
thanks a lot for the quick release!

PR #834, opened 2026-07-17
Pin release workflow actions

longlho:
## Why

The release workflow cannot run in organizations that require actions to be pinned to immutable commit SHAs. Its Intel macOS build also targets `macos-13`, which GitHub retired in December 2025.

## What changed

- Pin every third-party action to the commit behind its current major-version tag.
- Move the Intel build to `macos-15-intel`.

## Verification

Dispatched the workflow from a fork with the same patch. All five binaries built and uploaded successfully, and the npm packaging job completed.

Validation run: https://github.com/perplexityai/bazel-watcher/actions/runs/29595791430

Runner retirement: https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,7 +32,7 @@ jobs:
 
           - name: darwin_amd64
             artifact: ibazel_darwin_amd64
-            os: macos-13
+            os: macos-15-intel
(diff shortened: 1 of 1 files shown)


longlho:
thanks @achew22 !

PR #835, opened 2026-07-17
Support multiple FSEvents watch roots

### LINKED ISSUE #819: fsevents watcher fails with 'could not find common root of directories' on macOS
## Description

When running `ibazel build` on macOS, the fsevents watcher logs:

```
iBazel [3:17PM]: Error(s) updating watch list:
 could not find common root of directories
```

File watching still appears to work (builds trigger on changes), but the error is confusing and suggests something is broken.

## Cause

`findCommonRoot` in `internal/ibazel/fswatcher/fsevents/fsevents.go` computes the common prefix of all directories to watch. When the list includes paths with no shared root (e.g. workspace source files and external dependency paths), `rootLength` drops to 0 and the function returns an error.

https://github.com/bazelbuild/bazel-watcher/blob/main/internal/ibazel/fswatcher/fsevents/fsevents.go#L101-L123

## Reproduction

```bash
bazel run @multitool//tools/ibazel -- build //some/target:with_external_deps
```

On macOS with any target that has dependencies resolved from external repositories.

## Environment

- macOS (darwin arm64)
- ibazel v0.25.3
- Bazel 8.3.1 (Aspect fork)

## Suggestion

The function could either:
- Fall back to watching `/` (or each directory individually) instead of erroring
- Use multiple fsevents streams for disjoint directory trees

longlho:
## Summary

On macOS, iBazel collapsed every watched directory into one FSEvents root. When a workspace and a local or overridden repository lived under different top-level directories, no common root existed, so updating the watch list failed.

Group watched directories by top-level tree and compute one recursive watch root per group. This preserves the existing protection against FSEvents' 4,096-path limit without falling back to watching `/`.

Add a Darwin E2E with the Bazel workspace under `/private` and an overridden repository under `/Users`; changing the external source must trigger another run.

Fixes #819.

## Test

- `bazel test //... --test_output=errors`

### DIFF
--- a/internal/e2e/local_repository/local_repository_test.go
+++ b/internal/e2e/local_repository/local_repository_test.go
@@ -3,10 +3,11 @@ package simple
 import (
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
+	"os/user"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


longlho:
gentle ping @achew22  😄

PR #838, opened 2026-08-06
Start notification-mode runs before watch discovery

longlho:
## Why

`ibazel run` discovers the full transitive watch graph before starting the target. In large workspaces, this can delay a notification-mode dev server by tens of seconds even though the process can stay alive while iBazel installs watchers.

## What

- inspect the target tags up front, then start `ibazel_notify_changes` targets before full watch discovery
- install watchers after startup and send a synthetic graph change for a catch-up build
- support both legacy and v1 notification protocols
- preserve query-first behavior for restart-mode targets, builds, tests, and coverage

## Verification

The notification example served before watch discovery completed, received the catch-up build without restarting, then rebuilt normally after a source edit.

### DIFF
--- a/internal/ibazel/ibazel.go
+++ b/internal/ibazel/ibazel.go
@@ -200,18 +200,9 @@ func (i *IBazel) Cleanup() {
 
 func (i *IBazel) targetDecider(target string, rule *blaze_query.Rule) {
 	for _, l := range i.lifecycleListeners {
-		// TODO: As the name implies, it would be good to use this to make a
-		// determination about if future events should be routed to this listener.
-		// Why not do it now?
-		// Right now I don't track which file is associated with the end target. I
-		// just query for a list of all files that are rdeps of any target that is
-		// in the list of targets to build/test/run (although run can only have 1).
-		// Since I don't have that mapping right now the information doesn't
-		// presently exist to implement this properly. Additionally, since querying
-		// is currently in the critical path for getting something the user cares
-		// about on screen, I'm not sure that it is wise to do this in the first
-		// pass. It might be worth triggering the user action, launching their thing
-		// and then running a background thread to access the data.
+		// TODO: Use a target-to-file mapping to decide which listeners receive
+		// future events. Watch queries currently return the union of files for all
+		// requested targets and discard that ownership information.
 		l.TargetDecider(rule)
 	}
 }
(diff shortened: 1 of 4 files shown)


longlho:
this improved our cold start for web specifically by like 40%

longlho:
gentle ping @achew22 pls pls :)

longlho:
im also more than happy to help maintain this project as well

longlho:
gentle ping @achew22

internal/ibazel/ibazel.go
```diff
@@ -263,28 +257,28 @@ func (i *IBazel) setup() error {
...
 func (i *IBazel) Run(target string, args []string) error {
 	i.args = args
-	return i.loop("run", i.run, []string{target})
+	return i.loop("run", i.run, []string{target}, i.prepareRun(target))
 }
 
 // Build the specified targets in the IBazel loop.
 func (i *IBazel) Build(targets ...string) error {
-	return i.loop("build", i.build, targets)
+	return i.loop("build", i.build, targets, QUERY)
```
achew22: Why is passing the state into a call to `i.loop` superior to setting `i.state`?

internal/ibazel/ibazel.go
```diff
@@ -84,6 +84,9 @@ type IBazel struct {
 
 	lifecycleListeners []Lifecycle
 	pendingChanges     []command.Change
+	runCommandNotifies bool
```
achew22: Generally the way I would approach this would not be to add a bunch of boolean variables but instead to define enums (like I did here)

https://github.com/longlho/bazel-watcher/blob/3517bd33ddaf023fe6e25893e989fa8e39c12c9d/internal/ibazel/ibazel.go#L54-L61

and use them to define the modes of operation. WDYT about a new internal type `executionMode` and passing that around with a name for the thing you're interested in, then a switch on the enum in the places where you want to differ the behavior with the `default` being the old behavior?

achew22 (changes requested):
Can you also include some e2e tests that confirm the new functionality?

internal/ibazel/ibazel.go
longlho: `i.state` makes sense! done

internal/ibazel/ibazel.go
longlho: yup sgtm!

achew22 (approved):

